PR #1, opened 2020-06-05
Example messages

aaltat:
Does not work and now idea how do that with TS. I looked at Python and.node examples, but could not figure out a way to:
1) Read send parameters by one parameter at the time
1) Send response that contains data


src/index.ts
```diff
@@ -2,19 +2,21 @@ import { IPlaywrightServer, PlaywrightService } from './generated/playwright_grp
...
         await this.browser.close();
         callback(null, new Empty());
     }
 
-    async openBrowser(call: ServerUnaryCall<Empty>, callback: sendUnaryData<Empty>): Promise<void> {
+    async openBrowser(call: ServerUnaryCall<KeywordRequest>, callback: sendUnaryData<Response>): Promise<void> {
+        console.log("Open browser: " + call.request)
         this.browser = await firefox.launch({headless:false});
-        callback(null, new Empty());
+        callback(null, {rc: 'foo'});
```
mkorpela, line 19:
```
-        callback(null, {rc: 'foo'});
+        const response = new Response();
+        response.setRc('foo');
+        callback(null, response);
```

mkorpela:
Seems to work after the change I made.
```
Listening on 50051
Open browser: firefox
Close browser: alias 1,foo
Open browser: chrome
Close browser: alias 1,foo
Open browser: firefox
Close browser: alias 1,foo
Open browser: firefox
Close browser: alias 1,foo
Open browser: firefox
Close browser: alias 1,foo
```

aaltat:
Thanks for the pointers. Not ready to merge but somewhat more closer to ready.

mkorpela:
I think the error handling part should be done with grpc error handling mechanism https://grpc.io/docs/guides/error/

aaltat:
Even better with this https://cloud.google.com/apis/design/errors#error_model

What about logging?

PR #15, opened 2020-06-13
Max line leght to 100 charactes

aaltat:
Lets discuss about this one.


mkorpela:
I would go directly to 180 chars 😃

PR #50, opened 2020-06-23
Add list support

xylix:
- [x] Fix getting list selections
- [x] fix selecting from dropdown list (problem is reproduced in test Select From List)


xylix:
Current problem with getting this working seems to be mostly related to how we parse / pass / playwright treats our `{label: "Email"} matcher. 

However this playwright pen https://try.playwright.tech/?s=z3h6c works, so it can't be only on PW side. I also tried coding the selector and matcher in .ts side to ensure it's not a problem with variable passing or something weird like that.

It could also be related to my select content fetching method https://github.com/MarketSquare/robotframework-browser/pull/50/files#diff-8de88b7bcd474a2cd0ffd131cb81e50dR173 but after adding `Small select from list` test this seems unlikely since it doesn't do any list entry validation, only tries to mutate the dropdown once.

Docs: 
PW api doc and source code
https://github.com/microsoft/playwright/blob/master/docs/api.md#pageselectoptionselector-values-options
https://github.com/microsoft/playwright/blob/1865c5685a34a2e91ace07e08aee7ef15261406c/src/dom.ts#L408

Issue #55, opened 2020-06-27
Assertion engine


mkorpela:
**Is your feature request related to a problem? Please describe.**

There must be two style of keywords: asserting and value returning.

This may cause a keyword number explosion like with SeleniumLibrary
` Element Should Be Disabled, Element Should Be Enabled, Element Should Be Focused, Element Should Be Visible, Element Should Contain, Element Should Not Be Visible, Element Should Not Contain`
And the actual value returning `Get WebElement`. In this more extreme example that would require 8 keywords.

Value returning keywords are needed for automation, more complex assertions and if the value is later needed as an input to the system under test.

Using value returning assertion has currently fairly verbose syntax in Robot Framework:

     ${actualTitle}=  Get Title
     Should Be Equal  ${actualTitle}  Expected title

Keyword number explosion should be prevented. To make Robot Framework testing tool as good as there can be we should also aim for assertions that are easier to write in Robot Framework than in other languages.

**Describe the solution you'd like**
Lets built value getting keywords with assertion engine as the last arguments.
This might contain equals operator `== Expected title`, unequality `!= Unexpected title`, reqexp matching `=~ (\w+)xpected title` etc.

Syntax would be very short and in my opinion clean:

     Get Title  ==  Expected title

If done correctly, I believe this will result in a less verbose, less complex and more convinient way of doing validators.

mkorpela:
Done after https://github.com/MarketSquare/robotframework-browser/commit/e28f12b327e4e307df27e22605dbdcb845979a01

PR #57, opened 2020-06-28
AssertionEngine get url poc

mkorpela:
This is a sketch of AssertionEngine and it works.
I used it for `Get URL` keyword.

Unfortunately I also detected that https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#argument-types does not fully support enums. This might be a bug also in RF as I tried this for `Enum("AssertionOperator", "== !=")` or I did not really understand how to get it work.


mkorpela:
Got Enum working. The key is *not* to make it Optional. So the type and the default value must be from that specific enum.

mkorpela:
Linked to Issue #55

xylix:
I build a bit on this (added two more keywords, moved it to it's own class and added an optional message argument to enable more contextual info on the assertion failures). The syntax of the robot use looks quite good compared to the former.

This thing is something that will need some work before we want to release this:
- [ ] The keyword documentation's type signatures are not very detailed about how the enum works, and it doesn't automatically format it nicely.
Screenshot 2020-06-29 at 9 50 10

PR #60, opened 2020-06-30
Added more assertion operators and more robust with operators.

Snooz82:
Hey Browser-Team,
Hi @mkorpela Mikko,

i made some changes to the assertion engine.

Changed the `matches` operator to re.search, because otherwise it must start with this regex. 
The rest Regex BuiltIn Keywords do also match also partially. If a full match is needed you can do `^start_to_end$`.
This should be much more intuitive.

Also added `^=` / `should start with` and `$=` / `should end with`

also added `*=` as an alternative for `contains` 

Added `Should Be` and `Should Not Be` as an alternative for `==` and `!=`

Also refactored the code a bit, that it is more robust against human mistakes:
I changed `if operator.name == "<" and value >= expected:` to `if operator.name == "<" and not value < expected:`

Hope this contriburtion is welcome.
added some utest.

Cheers
René


mkorpela:
There is a lot of good stuff here. Unfortunately this still requires some changes.

**1. Shorthand operators**
I chose `==` and `!=` because they are universally used in programming languages for equality and inequality.
`^=`, `$=` and `*=` seem self-invented.
`*=` collides with https://docs.python.org/3/library/operator.html#operator.imul.
`^=` collides with https://docs.python.org/3/library/operator.html#operator.ixor.
So best to use words here. I propose `starts` and `ends`. Only common shorthands operators for now.

**2. Enum conversion**
Let's make Robot Framework do the work of converting values. No string things.

**3. No generated files**
Nope. These are only changed when Python <-> TypeScript protocol changes.

**4. One word operators**
We might need to make all assertion_operators one word. Enum does not seem to allow spaces.

Browser/assertion_engine.py
```diff
@@ -3,32 +3,83 @@
...
-    if operator.name == "<=" and value > expected:
+def _normalize(operator: str):
+    return operator.lower().replace('_', '').replace(' ', '')
+
+
+def verify_assertion(value: Any,
+                     operator: AssertionOperator,
+                     expected, message=""):
+
+    if isinstance(operator, str):
```
xylix, line 19: Could we get higher readability just by defining enum aliases for all the values that have multiple labels and matching by that below?

Browser/assertion_engine.py
```diff
@@ -3,32 +3,83 @@
...
+            operator = AssertionOperator["!="]
+        elif _normalize(operator) == "shouldstartwith":
+            operator = AssertionOperator["^="]
+        elif _normalize(operator) == "shouldendwith":
+            operator = AssertionOperator["$="]
+        else:
+            raise SyntaxError(f"{message}"
+                              f" `{operator}` is no valid assertion operator")
+
+    if operator.name in ["==", "should_be"] and not value == expected:
```
xylix, line 38: Is this `in` check necessary after we normalize and convert to a single possible value above?

Browser/generated/playwright_pb2_grpc.py
```diff
@@ -1,11 +1,12 @@
 # Generated by the gRPC Python protocol compiler plugin. DO NOT EDIT!
+"""Client and server classes corresponding to protobuf-defined services."""
```
xylix, line 2: This would need to be added to the `playwright.proto` for it to be persistent across protobuf re-generations, otherwise it will just disappear. I don't actually know if the proto file supports adding a doc comment to the top of the generated files.

xylix: Actually were these (and the changes in playwright_pb2.py) generated automatically from an updated version of the protobuf library? If so, It'd be good to also update requirements.txt with the newer grpc version.

Browser/assertion_engine.py
Snooz82: Enum Aliases are only possible when you define the enum as class, but then the members `==` and `>` are not possible.
So no aliases.

Also we do not need the enum really. it is just needed for future autocompletion.

Browser/assertion_engine.py
Snooz82: the in is because we do not have enum aliases and two enums are matching the same behaviour.
These are not strings, but two members of the enum.

Browser/generated/playwright_pb2_grpc.py
Snooz82: ¯\_(ツ)_/¯

Snooz82:
I can delete all these string matching except the enum available things and the normalization of these.
i think we should be robust against case.

The operators are not selv inveted, but they are the operators available in CSS.
Because we are not in python, but on Robot and in Web-Testing i thought these operators would be ok.
https://www.w3schools.com/cssref/css_selectors.asp

because the "multiple" words operators can not be part of the enum but they are very often used in robot, i thing, users will feel comfortable with them. LSP and enum support will probably not come before 2021. 
So as long as they are not supported, we can also support these via string matching.
the enum Conversion can also be modified, that pekka could do normalized member matching when converting.
like between method names and robot keywords.`should_be_equal` == `Should Be Equal`



What do you mean with 3?
Generation... should these files be in gitignore?

Browser/assertion_engine.py
aaltat: Could one pick `AssertionOperator` from a dictionary to improve readability? Example if one have: 
```
operators = {'contains': AssertionOperator["*="], 'shouldcontain': AssertionOperator["*="] ... }
operator = operators.get(operator)
```
The if madness would be gone and it could be easier to read.

xylix:
Yes we should try to be robust against casing. But I agree with Mikko that it should be through Robot handling the input casing correctly and passing us the right converted Enum.

I think the point about the operator naming is valid since they are valid CSS matchers. We should document them in the `Assertions` part of the KW documentation though (it's generated based on __init__.py:s docstrings. I also think we should try to make use of enum aliases here.

About the generated code, it's in the repo so users who install through pip etc don't need to generate the protocols again. I don't think it's necessary to .gitignore, but they should only be changed when the protocol definition changes. If you didn't change anything about the playwright.proto then it's probably a version disparity between GRPC library versions or a platform difference. We could aim to remove them in the future, but it's probably easiest to merge if you just remove them from the PR for now.

PR #115, opened 2020-07-15
More stable playwright object id's, Close All browsers and better switch context/browser behaviour

xylix:


atest/test/01_Browser_Management/playwright_state.robot
```diff
@@ -36,37 +36,37 @@ Create Browser does not create a context
...
-    Pass Execution    Switch Browser doesn't work yet
+    ${first_browser}    Create Browser    chromium
     Create Page Login
-    Create Browser    firefox
+    ${second_browser}    Create Browser    firefox
     Create Page Form
-    Switch Browser    0
+    ${after_switch}    Switch Browser    ${first_browser}
+    Should Be Equal As Numbers    ${first_browser}    ${after_switch}
     Get Title    matches    (?i)login
```
xylix, line 60: This doesn't pass even though the above row succeeds.

PR #149, opened 2020-07-22
Enchancement/python usage

xylix:


Browser/utils/logger.py
```diff
@@ -0,0 +1,19 @@
+from robot.api import logger  # type: ignore
+from robot.libraries.BuiltIn import EXECUTION_CONTEXTS  # type: ignore
+
+if not EXECUTION_CONTEXTS.current:
```
mkorpela, line 4: Is there an execution_context at the time that RF imports this module?

PR #161, opened 2020-07-25
Fix the Wait For Response acceptance test

idxn:


mkorpela:
Seems the implementations are still buggy. Thank you for the contribution!

PR #170, opened 2020-07-29
Raise exception at library import if node not accessible.

xylix:
WIP: print message to STDout instead of only syslog if possible


xylix:
Done in 15f53a2f117be4f2f8e596c5db402ac55e48d86e

PR #196, opened 2020-08-02
Autoclosing

mkorpela:
This is first version of Autoclose feature.
During this time also detected few test cases that were accidentally passing.

### DIFF
--- a/node/playwright-wrapper/playwright-state.ts
+++ b/node/playwright-wrapper/playwright-state.ts
@@ -100,24 +100,36 @@ export class PlaywrightState {
         this.activeBrowser = undefined;
     }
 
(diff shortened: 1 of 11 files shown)


node/playwright-wrapper/playwright-state.ts
```diff
@@ -100,24 +100,36 @@ export class PlaywrightState {
...
+                                            title: await page.p.title(),
+                                            url: page.p.url(),
+                                            id: page.index,
+                                        };
+                                    }),
+                                ),
+                            };
+                        }),
+                    ),
+                    activePage: browser.page?.index,
```
mkorpela, line 127: these are still wrong.

PR #207, opened 2020-08-04
Use proper logger in cookies

aaltat:


node/playwright-wrapper/cookie.ts
```diff
@@ -19,7 +22,7 @@ interface CookieData {
 
 export async function getCookies(callback: sendUnaryData<Response.String>, context?: BrowserContext) {
     const allCookies = await invokeOnContext(context, callback, 'cookies');
-    console.log(allCookies);
+    logger.info('Cookies: ' + JSON.stringify(allCookies));
```
xylix, line 25: https://getpino.io/#/docs/api?id=mergingobject
I think the best way to log these with pino is to do `logger.info( { 'cookies': allCookies }) and then it will be in a nice parseable format in the logs.

PR #224, opened 2020-08-06
Use Pino in TS side logger

aaltat:


xylix:
Rebasing master should fix the CI problem now that master has working PW 1.3.0

PR #251, opened 2020-08-13
Replace make with invoke

yanne:


yanne:
This is now passing the CI and local dev seems fine.

Can you proofread the docs, please?

Further development would be to implement some kind of timestamp checking in vein of make, since now all actions just always run.

yanne:
Added some rudimentary change checks to prevent building in vain.

xylix:
I tried a clean clone and the commands in a venv, worked well. We could maybe have a couple line entry script (probably just in bash or python?) to initialize a virtualenv + install invoke so users can directly start running the commands?

Should `inv atest` depend on `inv build` like we previously did with the Makefile?

Otherwise looks good to  merge

aaltat:
Document looks fine.

tasks.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+@task
+def clean_atest(c):
+    if atest_output.exists():
+        shutil.rmtree(atest_output)
+
+
+@task(clean_atest)
+def atest(c):
```
xylix, line 139: Should this depend on the build task (like the makefile used to) so it's always run on newest changes?

yanne: Good point, I'll add it here

PR #252, opened 2020-08-13
updated documentation

Finalrykku:


Finalrykku:
I correct these first. Interesting.. How could I pushed this in the first place then.

PR #271, opened 2020-08-17
Add shell=True when installing npm packages.

aaltat:
For windows.


aaltat:
Need a better solution that this one. Closing.

PR #274, opened 2020-08-17
Added tests for click with options and add styles

Finalrykku:
All tests of Click with options fail.
Add styles is fine


Finalrykku:
Since there is an update (https://github.com/MarketSquare/robotframework-browser/pull/278) on some tests of this commit, I close this pull request.

PR #286, opened 2020-08-19
Change page / context / browser indices to stable uuids

xylix:
Not mergeable yet, we need `Switch Page  LATEST` feature to avoid being unable to switch into pages that have been opened by page.


xylix:
TODO:
- [ ] Find a way to diff the browser page diffs ignoring the specific ids. Currently skipped

PR #314, opened 2020-08-23
Retry assertions until

mkorpela:
Mechanism to timeout assertions faster than with global timeout.


Snooz82:
i would prefer `retry_assertions_for` because this would fit better in our other keywords and is better english.

`retry_assertions_for=2 seconds`

aaltat:
I am fine with the argument and happy with the pr

PR #338, opened 2020-08-26
Mostly working ALL enum support

### LINKED ISSUE #284: Switch / Close keyword improvements
- [x] Support SeleniumLibrary style `NEW` argument in Switch / Close keywords
  - Should this technically be the second to last one since the last one is _current_?
  - Add a new data structure that contains all the pages / contexts besides the stack. Stack if for tracking the usage history
- [x] Document the options at a single place or keyword specifically?
- [x] Implement `ALL` enum functionality

xylix:
Closes #284

For pages there's still some bugs left to squash.

Also design input required: Now that we don't anymore have numerical indexes, workflows like `Close Page  1  ALL  ALL` wouldn't make sense anymore since all indices are unique to the specific pages. Flows like `Close Page  ALL  ALL  CURRENT` could still be supported but closing all pages in all contexts without closing the contexts seems a bit nonsensical.

### DIFF
--- a/Browser/browser.py
+++ b/Browser/browser.py
@@ -626,9 +626,7 @@ class Browser(DynamicCore):
         ]
         new_page_ids = [p for p in pages_after if p not in pages_before]
         for page_id, ctx_id in new_page_ids:
-            self._playwright_state.switch_context(ctx_id)
-            self._playwright_state.switch_page(page_id)
-            self._playwright_state.close_page()
+            self._playwright_state.close_page(page_id, ctx_id)
         # try to set active page and context back to right place.
         for browser in catalog_after:
             if browser["activeBrowser"]:
--- a/Browser/keywords/getters.py
+++ b/Browser/keywords/getters.py
@@ -440,11 +440,11 @@ class Getters(LibraryComponent):
 
             On the root level the data contains a list of open browsers.
 
-            Browser: ``{type: Literal['chromium', 'firefox', 'webkit'], 'id': int, contexts: List[Context]}``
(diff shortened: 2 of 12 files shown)


aaltat:
I guess the most common use case is to close all pages and the currently active context. Like I have two context open, because I want to test the webapp with two different logins/pages. But then I want to close all pages and the context from the first user, but leave second context and pages open because I want to continue testing with that user. Also I do not want close the browser. Something like this: 
` | Close Page | ALL | CURRENT | None |`

xylix:
If I understood your description correctly, wouldn't it make more sense to just `Close Context  CURRENT` in that case?

aaltat:
Hmmm, yes, you right, using `| Close Context | CURRENT |` is better.

mkorpela:
Some notes from daily:

This does not make sense now: `Switch Page   someid   ALL`.

`Close Browser  ALL`
`Close Page  ALL ?? ??? ??` .. this should close all pages in the current context?

`Keyword  page  context  browser`
`Keyword  ALL` => `Keyword  ALL  CURRENT CURRENT`
`Keyword` => `Keyword  CURRENT CURRENT CURRENT`

`Close Page`, closes CURRENT page, switch to previous.
`Close Page  uuid`, does not switch context nor browser unless uuid = CURRENT.

`Switch Page  uuid`, switch page, context and browser.

`${page_id}=   New Page`
`${popup_id}=   Switch page   NEW`
`${popup_id}=   Get Current Page Id`

`Get Browser Catalog   then   value[dsksdjsa]`

PR #345, opened 2020-08-27
GetBrowserCatalog to use Response.Json

aaltat:


node/playwright-wrapper/playwright-state.ts
```diff
@@ -3,7 +3,7 @@ import { ServerUnaryCall, sendUnaryData } from 'grpc';
 import { v4 as uuidv4 } from 'uuid';
 
 import { Request, Response } from './generated/playwright_pb';
-import { emptyWithLog, stringResponse } from './response-util';
+import { emptyWithLog, jsonResponse, stringResponse } from './response-util';
```
Snooz82, line 6: I think here missed to delete the string response

PR #404, opened 2020-09-25
Get Elements returns empty list when element is not found

### LINKED ISSUE #366: Get Elements should not raise timeout exception when no elements found. It should return an empty list.
Selectors can match multiple targets. Currently Get element just "randomly" returns the first match.
This might not be the correct thing.
Get Element is used so that further manipulations or validations can be made to that element. Selector might match 0, 1 or N elements.

Replace Get Element with Get Elements that returns a list of matches. Also have assertion engine there for easy manipulation of that list such as `Get Elements   selector   then   value[0]`

Get Element might also raise an error when multiple matches.

aaltat:
This feels ugly but could not understand how to get it working in TS side. 

Fixes #366

### DIFF
--- a/Browser/keywords/getters.py
+++ b/Browser/keywords/getters.py
@@ -508,9 +508,16 @@ class Getters(LibraryComponent):
 
         ``selector`` <str> Selector from which shall be retrieved. **Required**
         """
-        with self.playwright.grpc_channel() as stub:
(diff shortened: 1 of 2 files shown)


mkorpela:
This most likely just waits for the timeout. So it does not fix.

Issue #411, opened 2020-10-11
Option to stop execution on failure and keep browser and page open


mkorpela:
**Is your feature request related to a problem? Please describe.**
This has now been wished multiple times.

Can i prevent automatic closing of a browser?
I would have to investigate the state of the browser when a testcase i am working on fails. but the browser closes automatically when finished

**Describe the solution you'd like**
Add flag to stop execution on failure to open browser.

mkorpela:
Fixed in #412

PR #412, opened 2020-10-11
Topic/pause on failure

mkorpela:
- leave everything open and paused when failure happens after `Open Browser`
- default to *True*

Strong opinion about Open Browser being for development.


mkorpela:
https://github.com/MarketSquare/robotframework-browser/issues/411

PR #417, opened 2020-10-13
Use deltatime

aaltat:
Use RF deltatime argument type to convert arguments which takes time to
seconds.


aaltat:
Not yet readey, in the morning I did spot things that I need to look again.

PR #420, opened 2020-10-15
add video options to new context

mkorpela:


mkorpela:
I believe @aaltat has now implemented the real thing.

PR #484, opened 2020-11-06
Secrets with env variables

mkorpela:
Use env variables to hold secrets in `Type Secret` and `Fill Secret` keywords.

Syntax is the following: `Type Secret  selector  %MY_ENV_VARIABLE`

Env variable name is prefixed with `%`.

### DIFF
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -119,7 +119,8 @@ class Interaction(LibraryComponent):
         delay: timedelta = timedelta(seconds=0),
         clear: bool = True,
     ):
-        """Types the given secret from ``environment_variable`` into the found by ``selector``.
+        """Types the given secret from ``environment_variable`` into the text field
+        found by ``selector``.
 
         This keyword does not log secret in Robot Framework logs.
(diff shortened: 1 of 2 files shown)


Browser/keywords/interaction.py
```diff
@@ -138,48 +142,45 @@ def type_secret(
...
     @keyword(tags=["Setter", "PageContent"])
-    def fill_secret(self, selector: str, secret: str):
-        """Fills the given ``secret`` into the text field found by ``selector``.
+    def fill_secret(self, selector: str, environment_variable: str):
+        """Fills the given secret from ``environment_variable`` into the
+        text field found by ``selector``.
 
-        The difference to `Fill Text` is that this keyword does not log secret
-        on the INFO level. If ``enable_playwright_debug`` is enabled in the library
+        This keyword does not log secret in Robot Framework logs.
```
aaltat, line 155: This is not correct from users point of view. Although the keyword does not log it, RF will log it, if log level is high enough. But for normal users it is not possible to see the difference between the library keyword and RF core.

PR #485, opened 2020-11-06
Added note about upcoming changes in 3.0 release.

aaltat:


aaltat:
Not doing like this after discussion.

PR #487, opened 2020-11-06
Fix Get Cookie OSError

### LINKED ISSUE #479: Get Cookie fails if cookie does not have expiry date and return_type=dict.
Get Cookie fails if cookie does not have expiry date and return_type=dict. Example like this: 

image

Keyword should succeed if expiry date does not exist.

aaltat:
Fixes #479


aaltat:
Looked from CI logs, this fixes the problem.

PR #519, opened 2020-11-23
Add conda installation instructions

martinRenou:


aaltat:
Thank you for the contribution, but I think we are not going to accept this one. There are many ways to install Python packages and our project is not going to list them all. Also if we list an installation way, we state that we provide support for that installation method (meaning we have CI infra, we monitor it and so on) and that maintenance task we are not willing to accept.

xylix:
I think the conda package https://github.com/conda-forge/robotframework-browser-feedstock (since it is going to be used internally by Robocorp) will be maintained by me and / or @martinRenou . I could even set up test CI for it.

But it's true that for normal RF users conda is quite uncommon use case. So at least for now adding the instructions would be a bit pointless

martinRenou:
> I could even set up test CI for it.

I added a test on the recipe that checks that the `Browser` Python module is correctly installed https://github.com/conda-forge/robotframework-browser-feedstock/blob/master/recipe/meta.yaml#L39-L41

It's a small test though, I wanted to run a `test.robot` file making use of the library but faced an issue with libx11 not being available on the CI. That could be done I guess.

We could also add some tests checking that the `node_modules` directory is correctly generated in the right place (because you depend on it if I'm correct).

> but I think we are not going to accept this one. There are many ways to install Python packages and our project is not going to list them all

To be fair, conda is quite popular now and used by many packages. It is way safer to install a package with conda than it is with pip, because conda actually does dependencies solving while pip doesn't.

> that maintenance task we are not willing to accept

If dependencies are not modified, the maintenance is fairly small with conda-forge. There is a bot that opens PRs to update the conda package when a release has been made on PyPi: https://github.com/conda-forge/robotframework-browser-feedstock/pull/1

aaltat:
> > but I think we are not going to accept this one. There are many ways to install Python packages and our project is not going to list them all
> 
> To be fair, conda is quite popular now and used by many packages. It is way safer to install a package with conda than it is with pip, because conda actually does dependencies solving while pip doesn't.

Yes conda is popular, so is poetry. The point being is that we do not list all possible ways to install a Python package. Instead we say one and stick with it. In Python world, pip is still the defacto way, but users can freely choose their way of install.

> > that maintenance task we are not willing to accept
> 
> If dependencies are not modified, the maintenance is fairly small with conda-forge. There is a bot that opens PRs to update the conda package when a release has been made on PyPi: [conda-forge/robotframework-browser-feedstock#1](https://github.com/conda-forge/robotframework-browser-feedstock/pull/1)

Each activity has cost and even small cost tends to accumulate to big chunk for a long period. The C-Python core team thinks in decades when they decides to implement and support something. Although I am not thinking that far in the future, I am considering several years. For the past half year, which this project has been live, we have had few quite significant changes in the packaging and in related activities. Also I can see few others coming in the near future and for the long term, I do not have any idea. 

If someone else wants to support building a conda package, they are more that welcomed to do so .

PR #520, opened 2020-11-24
Add linting for from __future__ import annotations

xylix:
This `__future__` behaviour will be made mandatory in 3.10, and for packages wanting to support both older pythons and 3.10  at the same time using this is best practice.

The changed behaviour relates to later evaluation of type annotations, so for fixing the test failures work in gen_type_stubs, robot core or pythonlibcore might be necessary as well as work on our end. https://www.python.org/dev/peps/pep-0563/

### DIFF
--- a/Browser/.flake8
+++ b/Browser/.flake8
@@ -2,4 +2,5 @@
 exclude =
    __pycache__,
    generated,
-ignore = E501, W503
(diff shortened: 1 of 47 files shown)


mkorpela:
Too old.

PR #528, opened 2020-11-30
Argparse and rfbrowser settings saving

xylix:
Use argparse for CLI parsing, start adding support to save settings of rfbrowser init into file.

This enables installs where for example browser location is customzied at install time, without user needing to re-specify them at every import site.

TODO: 
- [ ] Get installation Path correctly in Browser/browser.py or abstract it into some other module.
- [ ] Enable flows like `rfbrowser init --browser=chromium` 
- [ ] Enable usage like `rfbrowser init --external-browser-executable=/chromium-path`


aaltat:
Arg parse is already in use, closing.

PR #551, opened 2020-12-07
WIP presenter mode

mkorpela:


janipalsamaki:
I would love to use this feature already today to be able to better demo bots!

The normal bot works under the covers and outputs stuff, but that is not the most spectacular show ever for anyone observing it. Being able to visualize the work would be great, even better if using some kind of toggle.

Any chance of getting this one merged as is? It would already provide value!

Browser/browser.py
```diff
@@ -599,6 +601,7 @@ def __init__(
         run_on_failure: str = "Take Screenshot",
         external_browser_executable: Optional[Dict[SupportedBrowsers, str]] = None,
         jsextension: Optional[str] = None,
+        enable_presenter_mode: bool = False,
```
aaltat, line 604: Instead of bool value, this could be better  be : `enable_presenter_mode: Optional[int] = None`
Then one could use the value to control how long the element highlighted.  Example like:
```
        if self.library.presenter_mode:
            self.hover(selector)
            self.library.highlight_elements(selector, duration=timedelta(seconds=self.library.presenter_mode))
```
And not to hard code the timeout in some fixed value.

Browser/keywords/interaction.py
```diff
@@ -297,6 +297,10 @@ def click(
         during the click, and then restores current modifiers back.
         If not specified, currently pressed modifiers are used.
         """
+        if self.library.presenter_mode:
+            self.hover(selector)
+            self.library.highlight_elements(selector, duration=timedelta(seconds=2))
+            sleep(2)
```
aaltat, line 303: Why we need sleep here and in the other places?

aaltat:
Sorry this slipped trough the cracks and no-one in the core team did pick up the ball.

Browser/browser.py
mkorpela: Hard coded value is easier to use. More thinking on our side with sensible default.

Browser/keywords/interaction.py
mkorpela: To make highlight happen before the click, text fill, typing, closing happens.

PR #571, opened 2020-12-12
Add args to run_on_failure

xylix:


aaltat:
There is already argument support, but in library import. Closing this PR.

PR #574, opened 2020-12-13
Feature/assert future dialog

emakaay:
@This code is not production ready.

The purpose of the pull request is to assert dialog text.

I created some tests, and code to perform this action. 
But don't understand how to get to the dialog message.

Perhaps someone can finish this pull request ? 

Hope it helps a bit ;)

### DIFF
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -619,7 +619,7 @@ class Interaction(LibraryComponent):
                     dialog_text,
                     assertion_operator,
                     assertion_expected,
-                    f"Asserting dialog text",
+                    "Asserting dialog text",
                 )
 
     @keyword(tags=("Setter", "PageContent"))
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -588,7 +588,8 @@ class Interaction(LibraryComponent):
         self,
         action: DialogAction,
         prompt_input: str = "",
-        expected_dialog_text: str = None
+        assertion_operator: Optional[AssertionOperator] = None,
+        assertion_expected: Any = None,
     ):
         """Handle next dialog on page with ``action``. Dialog can be any of alert,
         beforeunload, confirm or prompt.
@@ -597,8 +598,8 @@ class Interaction(LibraryComponent):
 
             ``prompt_input`` The value to enter into prompt. Only valid if
             ``action`` equals accept. Defaults to empty string.
-            ``expectedDialogText`` The text expected being on the dialog. Default to no checking
-            Optionally asserts the dialog message.
+            If ``assertion_operator`` is set and property is not found, ``value`` is ``None``
+            and Keyword does not fail. See `Get Attribute` for examples.
 
             See `Assertions` for further details for the assertion arguments. Defaults to None.
 
@@ -608,9 +609,18 @@ class Interaction(LibraryComponent):
             if prompt_input and action is not DialogAction.accept:
                 raise ValueError("prompt_input is only valid if action is 'accept'")
             response = stub.HandleAlert(
-                Request().AlertAction(alertAction=action.name, promptInput=prompt_input, expectedDialogText=expected_dialog_text)
+                Request().AlertAction(alertAction=action.name, promptInput=prompt_input)
             )
             logger.debug(response.log)
+            if assertion_operator is not None:
+                dialog_text = response.body
+                logger.debug(dialog_text)
+                verify_assertion(
+                    dialog_text,
+                    assertion_operator,
+                    assertion_expected,
+                    f"Asserting dialog text",
+                )
 
     @keyword(tags=("Setter", "PageContent"))
     def mouse_button(
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from time import sleep
 from typing import Any, Dict, Optional
 
+from ..assertion_engine import verify_assertion
 from ..base import LibraryComponent
 from ..generated.playwright_pb2 import Request
 from ..utils import (
@@ -41,7 +42,7 @@ from ..utils.data_types import (
     MouseButtonAction,
     MouseOptionsDict,
     ScrollBehavior,
-    SelectAttribute,
+    SelectAttribute, AssertionOperator,
 )
 
 NOT_FOUND = object()
@@ -50,11 +51,11 @@ NOT_FOUND = object()
 class Interaction(LibraryComponent):
     @keyword(tags=("Setter", "PageContent"))
     def type_text(
-        self,
-        selector: str,
-        text: str,
-        delay: timedelta = timedelta(seconds=0),
-        clear: bool = True,
+            self,
+            selector: str,
+            text: str,
+            delay: timedelta = timedelta(seconds=0),
+            clear: bool = True,
     ):
         """Types the given ``text`` into the text field found by ``selector``.
 
@@ -118,11 +119,11 @@ class Interaction(LibraryComponent):
 
     @keyword(tags=("Setter", "PageContent"))
     def type_secret(
-        self,
-        selector: str,
-        secret: str,
-        delay: timedelta = timedelta(seconds=0),
-        clear: bool = True,
+            self,
+            selector: str,
+            secret: str,
+            delay: timedelta = timedelta(seconds=0),
+            clear: bool = True,
     ):
         """Types the given secret from ``variable_name`` into the text field
         found by ``selector``.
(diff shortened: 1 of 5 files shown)


aaltat:
I understand you need, but for starters could you remove all changes which are done (most likely) your IDE, because it makes changes really noisy. Also this causes some Pep violations and CI therefore fails. Most likely the easiest solution is


1. Close your IDE or whatever you are using to write the code.
1. Follow the development environment setup [instructions](https://github.com/MarketSquare/robotframework-browser/blob/master/CONTRIBUTING.md#development-environment)
1. Run `inv lint` and then git add/commit/push your changes.

emakaay:
@aaltat 
Sorry for the problems i'm causing.
I have performed the inv lint command.

I would like to remove the pull request. But don't know how.

i'm not able to get the dialog message into the interaction.py.

aaltat:
I did not mean it in harsh way, sorry about that.

In general, there is usually feedback given, to both directions in a PR. Feedback can be related to many things, like way of coding style or in way feature is implemented, to almost anything under the sun. It is kind of discussion, with a delay, because most of the time, you and me are not present at the same time. And because we can not see each other, many things are lost in the translation.

But now the pull request looks pretty good and is not far from being accepted. Perhaps test or two and then all would have been good. So if you are interested, I can guide you trough the PR and get thing done.

emakaay:
Hello Tatu,

I didn't close the pull request because your comment was harsh.

I thought I had the skills to add the code to solve my issue (assert the
dialog message). But I'm in over my head.... And I don't want to add a sub
optimal solution.

Perhaps someone can finish this (or start over if that's easier)

This is the last thing I need to recommend using this great framework in
several projects.

Greets,
Erik

aaltat:
Your solution does look's pretty good. Also it is totally alright to be confused in a strange codebase, we are here to help and guide you trough the process, if you want.

emakaay:
Hello Tatu,

If you can guide me that would be helpful.
I have only one difficulty to solve this issue.

My understanding is that the following function from "interactions.ts", is
there to wait for a possible dialog. And if the dialog occurs, the dialog
is accepted or dismissed. I thought the line "wait invokeOnPage(page, 'on',
'dialog', fn)"  would return a Dialog object, but it doesn't. It returns a
Page object.

I would like to know how to get to the dialog message that is accepted or
dismissed. If I can get that, then the rest is already in place. I already
tested the stringResponse with Robot Test. The only problem is that not the
correct dialog message is passed through.

export async function handleAlert(request: Request.AlertAction, page?:
Page): Promise<Response.String> {
    const alertAction = request.getAlertaction() as 'accept' | 'dismiss';
    const promptInput = request.getPromptinput();
    const fn = async (dialog: Dialog) => {
        if (promptInput) await dialog[alertAction](promptInput);
        else await dialog[alertAction]();
    };
    await invokeOnPage(page, 'on', 'dialog', fn);
    return stringResponse('don't know were to get dialog message
from', 'Set event handler for next alert');

}"

greets,

Erik

aaltat:
Well, we have these generic invoke methods, in this case we use [invokeOnPage](https://github.com/MarketSquare/robotframework-browser/blob/master/node/playwright-wrapper/playwirght-invoke.ts#L59). That node side code is too clever for me, but lets ask @mkorpela or @xylix how that invokeOnPage could be changed so that it returns dialogue [message](https://playwright.dev/docs/api/class-dialog#dialogmessage)

emakaay:
@mkorpela 

Could you give me some pointers to get to the dialog message(as discussed above) ?
That would help me to finish the issue. 

Thanks in advance,
Erik

emakaay:
For some reason this is very hard.
I'm able to check the dialog message.
When the dialog message doesn't contain the expected text, I throw an Error.

This works, but I can't seem to handle the promise correctly (never used it before). ANd the playwright process stops because of this. Don't know how to proceed...

emakaay:
Still working on it

xylix:
```
export async function handleAlert(request: Request.AlertAction, page?:
Page): Promise<Response.String> {
    const alertAction = request.getAlertaction() as 'accept' | 'dismiss';
    const promptInput = request.getPromptinput();
    const fn = async (dialog: Dialog) => {
        const message = dialog.message();
        if (promptInput) await dialog[alertAction](promptInput);
        else await dialog[alertAction]();
        return message;
    };
    const message = await invokeOnPage(page, 'on', 'dialog', fn);
    return stringResponse(message, 'Set event handler for next alert');

}
```

You should be able to access the dialog message directly in the callback through the dialog. See my code example. Hopefully it helps. Feel free to ask more questions if it does not work or you have trouble with it :).

emakaay:
Hello Kerkko,

This very helpfull, but i'm still unable to get the dialog message

The type of message is 'undefined'. However I expected to get a type String.
Therefore if a add something to message like 'dummy', that is returned from
the function to the HandleAlert python method. This indicates that the
change from Respone.Empty to Response.String is working
So passing of the parameters is correct, but for some reason
dialog.message()

I also tried to get the message of the dialog after the wait. That did'n
help either. I don't understand why it won't work....

This is what i have now:

export async function handleAlert(request: Request.AlertAction, page?:
Page): Promise<Response.String> {
    const alertAction = request.getAlertaction() as 'accept' | 'dismiss';
    const promptInput = request.getPromptinput();
    const fn = async (dialog: Dialog) => {
        const message:String = dialog.message();
        if (promptInput) await dialog[alertAction](promptInput);
        else await dialog[alertAction]();
        return message
    };
    const dialogMessage = await invokeOnPage(page, 'on', 'dialog', fn);
    return stringResponse('dummy' + dialogMessage.name, 'Set event
handler for next alert');
}



with self.playwright.grpc_channel() as stub:
    if prompt_input and action is not DialogAction.accept:
        raise ValueError("prompt_input is only valid if action is 'accept'")
    response = stub.HandleAlert(
        Request().AlertAction(alertAction=action.name, promptInput=prompt_input)
    )
    logger.debug(response.log)
    if assertion_operator is not None:
        dialog_text = response.body
        logger.debug(dialog_text)
        verify_assertion(dialog_text, assertion_operator,
assertion_expected, f"Asserting dialog text")

Any ideas ??

Erik

emakaay:
Hello Kerkko,

I dug some more.... with adding some dumb console.log() lines.
I can see that the function(fn) can find the dialog message. But the
invokeOnPage doesn't return the string.
The result (from invokeOnPage) is undefined Object.... But again... i'm not
sure why ?

Any ideas about this ?
Erik

aaltat:
Without seeing you code in the PR itself it is hard to say what is wrong. Could you commit your changed and push it?

emakaay:
Hello  Tatu,

I have committed the changes.
Hope you can see were the problem is 😄

Erik

Browser/keywords/interaction.py
```diff
@@ -599,6 +612,15 @@ def handle_future_dialogs(self, action: DialogAction, prompt_input: str = ""):
...
             )
             logger.debug(response.log)
+            if assertion_operator is not None:
+                dialog_text = response.body
+                logger.debug(dialog_text)
+                verify_assertion(
+                    dialog_text,
+                    assertion_operator,
+                    assertion_expected,
+                    f"Asserting dialog text",
```
aaltat, line 622: Remove the `f` from the front of the string and CI will go further.

aaltat:
There is small pep violation, see the comment. Please fix it and lets see what is the next error to solve.

Browser/keywords/interaction.py
emakaay: Thanks !
That f character was not required anymore.

But I think the reason why I can't get this working is because the dialog text is not passed through.
When you look at the interaction.py code I log 2 things (the log, and the body.... The body should contain the dialog text. But it doesn't. That the real problem.... it isn't returned from the invokeOnPage function.

you can see this also in the robot log:

DEBUG	Test timeout 30 seconds active. 29.605 seconds left.	
DEBUG	Set event handler for next alert	
DEBUG		
DEBUG	Unknown error received: Asserting dialog text '' (str) should be 'Am an alert' (str)

The third line (corresponding with logging of the dialog text) is empty

aaltat: Ah, now I see the problem and remember how it works, sorry I did lead you to wrong direction. It due of the promises we use. 

When you have this test:
```
Dismiss Alert
    Handle Future Dialogs    action=dismiss
    Click    \#alerts
```
The `Handle Future Dialogs` sets a promise in the node side to handle the future alert. But when `Handle Future Dialogs` keyword is run, the alert is not yet there. The alert is only present when `Click` keyword is run and alert is handled in the background by playwright (Handle Future Dialogs).

But I think, means did not test, using promise should work:
```
Dismiss Alert
    ${promise}    Promise To    Handle Future Dialogs    action=dismiss
    Click    \#alerts
    ${text}=    Wait For  ${promise}
```
The point which makes me unsure that does promise and assertion engine work together. It might not work or then again it might work, no idea basically. Gut feeling says that it might not work.

xylix: My guess is that the problem is not the `invokeOnPage` , the problem is the dialog.text() returning undefined... are you sure the problem is at the invokeOnPage function?

aaltat: If it does not work, then you can return the text in the dialogue and assert the text by RF core keyword. It is not as coinvent as assertion engine, but could be still useful.

xylix: I don't think promise and the current event handler implementation will play nicely together. Might be necessary to change the ts side impl from a `await invokeOnPage(page, 'on', 'dialog', fn)` to `await invokeOnPage(page, 'waitFor', 'dialog', fn)`. But it might make sense to make that as a separate keyword even.

emakaay:
I still need the functionality, but can't build it myself, I'm sorry to say

PR #592, opened 2020-12-16
Add Node version limitation hint to the installation notes

simonmeggle:
Installation of Browser fails (and is not supported currently) on Node.JS version >= 15. 

This was already [discussed on Slack](https://robotframework.slack.com/archives/C015KB1QSDN/p1608043930266600). 

Please add this important fact to the installation instructions of browser.


xylix:
Thanks!

PR #635, opened 2021-01-06
Add support for suggestedFilename in Wait For Download keyword

### LINKED ISSUE #597: Wait For Download should also allow browser suggested filename
Currently with Wait For Download keyword it is only possible to download with a custom name (saveAs argument) or random guid.

The issue is similar with Playwright issue: https://github.com/microsoft/playwright/issues/2058 where the proposed solution was to use download.suggestedFilename()

The enhanced **Wait For Download** keyword, would allow for saveAs argument a special string to save with suggested filename, e.g. **Wait For Download**  _saveAs=DEFAULT_

As altenative I implemented a jsextension with a custom keyword **Wait For Download With File Name** that is returning a list with downloaded file path and suggested filename, that can be used in Robot Framework to rename the file. But obviously an integrated functionality in Browser library would be more elegant:

https://github.com/adrian-evo/robotframework-browser-extension

**Additional context**
In my web applications that I am testing (industrial web configurator for electrical motors and other industrial components), it is very important to check if the downloads respect certain naming conventions, e.g. like motors datasheets.

aaltat:
Fixes #597

### DIFF
--- a/Browser/keywords/waiter.py
+++ b/Browser/keywords/waiter.py
@@ -173,8 +173,10 @@ class Waiter(LibraryComponent):
             logger.info(response.log)
 
     @keyword(tags=("Wait", "BrowserControl"))
-    def wait_for_download(self, saveAs: str = "") -> str:
-        """Waits for next download event on page. Returns file path to downloaded file.
+    def wait_for_download(
+        self, saveAs: str = "", suggestedFilename: bool = False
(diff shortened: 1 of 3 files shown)


xylix (approved):
Like discussed in Slack, this could also be implmeneted in a backwards-incompatible but cleaner way (always return a dict), but getting it to users hands might be a higher priority.

PR #747, opened 2021-02-28
Topic/page listeners

mkorpela:
Recording page errors and console output.
This PR just transports them to the Python side in the Close Page phase.
They will be logged only if the close page is called explicitly.
Working on this by changing our auto closing and listener API to 3 so that we can impact test execution result and test model in _end_suite and _end_test.


VilleSalonen:
This is exactly what I was looking for. We have about 650 existing tests which are currently being run on SeleniumLibrary. I am in the progress of migrating them over to Browser. We have very similar functionality implemented currently with SeleniumLibrary so this PR would help us very much.

PR #795, opened 2021-03-11
Add upgrade instructions to manual

leeuwe:
It doesn't state yet that, to fully benefit from every next version of robotframework-browser, you need to run `rfbrowser init` again. This PR is to further dummy-proof that.


aaltat:
@all-contributors please add leeuwe for documentation

Issue #757, opened 2021-03-03
[Get Selected Options] keyword returns error "Error: page.$$eval: Evaluation failed: DOMException: Failed to execute 'evaluate' on 'Document': The string './/select[@name='brandCategory'] option' is not a valid XPath expression."


vuongvvv:
**Describe the bug**
[Get Selected Options] keyword returns error "Error: page.$$eval: Evaluation failed: DOMException: Failed to execute 'evaluate' on 'Document': The string './/select[@name='brandCategory'] option' is not a valid XPath expression."

**To Reproduce**
![2021-03-03_11-23-25](https://user-images.githubusercontent.com/23648562/109752635-565b9f00-7c13-11eb-9b0b-16a6113ab4ad.jpg)

1. `Get Selected Options    selector=//select[@name='brandCategory']           option_attribute =label        assertion_operator=equal          Food & Drink`

**Expected behavior**
"Get Selected Options" keyword should work with "//select[@name='brandCategory']" xpath

**Desktop:**
 - OS: MAC OS Catalina 10.15.7
 - Browser Chromium
 - Version 71.0.3542.0

aaltat:
The error message says that your xpath is `'.//select[@name='brandCategory'] option' ` and I agree that it's not valid xpath. Could you fix the xpath and try again.

vuongvvv:
- I tried with "//select[@name='brandCategory']" xpath -> Still got the same error

![Log_1](https://user-images.githubusercontent.com/23648562/109920245-98a4df00-7cec-11eb-8408-5f83d15b370b.jpg)
![Log_2](https://user-images.githubusercontent.com/23648562/109920255-9c386600-7cec-11eb-9e2f-3ee1fb90c10f.jpg)

aaltat:
This is actually a bug in Browser library side: https://github.com/MarketSquare/robotframework-browser/blob/bdf001ed3e33e0b5ea5fd146cf0da36dfcd0e766/node/playwright-wrapper/getters.ts#L49 We append  ` option` to the selector and therefore it fails.

emakaay:
I think issue #630 would be fixed with this also

PR #802, opened 2021-03-14
add get selected options xpath support

mkorpela:
Fixes https://github.com/MarketSquare/robotframework-browser/issues/757

### DIFF
--- a/atest/test/02_Content_Keywords/select_lists.robot
+++ b/atest/test/02_Content_Keywords/select_lists.robot
@@ -37,6 +37,10 @@ Get Selected Options
     Run Keyword And Expect Error    *
     ...    Get Selected Options    select[name=possible_channels]    label    ==    Email    Telephone    Direct mail
 
+Get Selected Options with xpath
+    ${selection}=    Get Selected Options    //html/body/form/table/tbody/tr[8]/td[2]/select    label    ==    Telephone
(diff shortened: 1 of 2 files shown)


emakaay:
@mkorpela 
Thanks for fixing issue 630.
The problem of incorrect xpath expression is gone.

But I still have a problem selecting the correct drop down options.

I have the following test:
Select Option And Verify selection    label    //*[@Id='dropdownid']/select     tobeselecteddropdownvalue

The keyword **select option and verify Selection** is copied from the tests
```*** Keywords ***
Select Option And Verify Selection
    [Arguments]    ${attribute}    ${list_id}    @{selection}
    Select Options By    ${list_id}    ${attribute}    @{selection}
    Get Selected Options    ${list_id}    ${attribute}    ==    @{selection} ```

However this still doesn't select the correct value. When I adjust the above keyword to:

```*** Keywords ***
Select Option And Verify Selection
    [Arguments]    ${attribute}    ${list_id}    @{selection}
    pause execution
    Select Options By    ${list_id}    ${attribute}    @{selection}
    pause execution
    Get Selected Options    ${list_id}    ${attribute}    ==    @{selection} ```

It will pass the test. The dropdown list is not fixed, but determined on runtime... I think there should be a waiting time when select option(s) are not present

PR #806, opened 2021-03-16
Update Browser-4.0.0.html

LDerikx:
Typo at the Click keyword. In arguments it says clickCount in the documentation click_count, clickCount is accepted in the keyword


aaltat:
Keyword documentation is automatically generated from the source code, therefore the bug should be fixed in the source code side.

Issue #818, opened 2021-03-19
Promise To Wait For Request continues too early


mkorpela:
**Describe the bug**
Wait For Request must have attached request listener before promise should be released.
This is now not happening. This will sometimes result in next action being executed before request listener is on.

https://github.com/MarketSquare/robotframework-browser/runs/2152226138?check_suite_focus=true

mkorpela:
Fixed

PR #820, opened 2021-03-20
Topic too slow promise

mkorpela:


mkorpela:
3 times green pass. Merging

mkorpela:
Fixes https://github.com/MarketSquare/robotframework-browser/issues/818

Issue #843, opened 2021-03-22
Upload file broken when used to handle more than a single file upload


xylix:
- Upload file "stacks up" event handlers when called multiple times
  - When there are multiple handlers they will race which one gets to upload it's own file
- no way to clear the existing upload file handlers  

https://robotframework.slack.com/archives/C015KB1QSDN/p1613976029002100
https://robotframework.slack.com/archives/C015KB1QSDN/p1615985793008300 

Proposed solutions:
- Make a single upload file call always clear after the file has been uploaded (e.g. instead of playwright `.on` use playwright `waitFor` event)
- Expose playwright Upload File keyword

PR #845, opened 2021-03-22
Xylix/fix upload file

xylix:
Deprecate broken `Upload File` add new `Promise to Upload File` keyword

Cat and dog seem to be failing. Presumably new `Promise to Upload File` or one of the tests leaks a promise that fails the other test.


xylix:
Fixes https://github.com/MarketSquare/robotframework-browser/issues/843

PR #882, opened 2021-03-28
finder based css selector creation

mkorpela:
This is insanely awesome. It is still just a proof of concept but it is just too good to be true.

Basically this test stops on that Click and you have 5 seconds to move your mouse to the element you want to click:
```RobotFramework
*** Settings ***
Library   Browser  enable_presenter_mode=true  timeout=30s

*** Test Cases ***
Example Test
   new browser  headless=false
   New Page  https://yourawesomesite.com
   Click  __DETECT__
   Sleep  5s
   Take Screenshot
```


emakaay:
👍

Issue #869, opened 2021-03-26
Support for Take Sreenshot Options


vincenzo-gasparo:
Hi, right now the supported arguments for the **Take Screenshot** keyword are: **filename, selector, fullPage**.

It would be very useful to support some of the other arguments present in the Node.js version of playwright, specifically:

- **quality <number>** The quality of the image, between 0-100. Not applicable to png images.
- **type <"png"|"jpeg">** Specify screenshot type, defaults to png.

Also, at the moment the keyword sometimes throws `page.screenshot: Timeout 10000ms exceeded` for big fullPage screenshots (I think because processing time), so maybe the **timeout** argument could be useful too.

- **timeout <number>** Maximum time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout. The default value can be changed by using the browserContext.setDefaultTimeout(timeout) or page.setDefaultTimeout(timeout) methods.
 
 [Link To Playwright API](https://playwright.dev/docs/api/class-page#pagescreenshotoptions)

mkorpela:
@all-contributions please add vincenzo-gasparo for code.

mkorpela:
@vincenzo-gasparo awesome work! Would like those enums also as Robot Framework will automatically typecheck strings from enums. But that can be another PR. This one is merged.

mkorpela:
@all-contributors please add vincenzo-gasparo for code

mkorpela:
@all-contributors please add vincenzo-gasparo for code

mkorpela:
Wtf @all-contributors, please autofix your bugs!!!

PR #895, opened 2021-04-02
Take screenshot arguments

vincenzo-gasparo:
fileType is now an Enum with png and jpeg options, added new arguments definition to python doc #869


aaltat:
Was close accident?

Issue #916, opened 2021-04-10
Wait For Navigation keyword should support regex and waitUntil


vincenzo-gasparo:
Wait For Navigation Keyword only support exact match for the URL, it would be nice to allow a partial match with a regex support.

Also the playwright API support a waitUntil argument. [Playwright API](https://playwright.dev/docs/api/class-page?_highlight=waitforna#pagewaitfornavigationoptions)

I thought about two ways to add the regex support:
- Add a **regex** boolean argument
- Use an enum to specify the type of the url

I prefer the first option since I don't think it would make sense to support the **function(URL):boolean** type

mkorpela:
@vincenzo-gasparo your PR is now merged. Is there additional tasks you would like to do related to this issue?

aaltat:
I am going to assume that there is not anything more left and I will close the issue. If there is more, please reply or open new issue.

PR #917, opened 2021-04-10
Added regex and wait_until args to wait_for_navigation

vincenzo-gasparo:
Added regex and wait_until args to wait_for_navigation #916


mkorpela:
Thanks for contributing! The boolean regex flag is a bit weak solution. Would there be away to make thecode directly identify if the selector is regex or not? Anyway this will be eventually approved!

vincenzo-gasparo:
Yes I thought about it, I tried enclosing regex between **/---/** but it's possible that someone want to match the **'/'** characters. What do you think about **//regextomatch//** or maybe **RegExp(regextomatch)**?

aaltat:
Would be possible to try with both? First with full match and if it fails then with regex?

vincenzo-gasparo:
> Would be possible to try with both? First with full match and if it fails then with regex?

Maybe trying both with a Promise.any? In order to avoid waiting the timeout two times

aaltat:
@all-contributors please add vincenzo-gasparo for code

aaltat:
Oh darn created https://github.com/all-contributors/all-contributors/issues/512

mkorpela (approved):

mkorpela:
One option could also be `regex=....` but then there is also wait for response (+ wait for request that nobody uses). It should follow the same pattern in my opinion - not like everyone having a different logic.

Snooz82:
Hi,

i do not get why `/regex/` is an issue?
A URL can not start with a slash character.

that argument is not a selector but a matcher.
Therefore i would not see `regex=` to be valid here.

`regex:<myregex>` could be an option, because it is a url.

 But the regex pattern with the slashes like
`/Log\s*in/i` Would be best, because it is same as in the text selector.

vincenzo-gasparo:
I added a check in order to verify if the url matches `^/(.*?)/([gimy]*)$'`, in this way it's possible to pass either a url or a regex in the form `/regex/flags` what do you think? Maybe it's a bit overkill

Snooz82:
No i think that is mo overkill. 👍🏼
But i think JS does know more flags.
flags: `i g s m y u`

You can maybe check what Playwright itself used in the text selector to detect Regex vs strings.

vincenzo-gasparo:
> No i think that is no overkill. 👍🏼
> But i think JS does know more flags.
> flags: `i g s m y u`
> 
> You can maybe check what Playwright itself used in the text selector to detect Regex vs strings.

I think that the playwright way of detecting regex is way simpler than I thought [Link](https://github.com/microsoft/playwright/blob/b62a43604196c6dfb03b4a3fb040e836e33c3f8f/src/server/injected/injectedScript.ts#L766), this should be the original code

vincenzo-gasparo:
I could simplify the code according to original pw implementation:

```export async function waitForNavigation(request: pb.Request.UrlOptions, page: Page): Promise<pb.Response.Empty> {
    const url = <string>request.getUrl()?.getUrl();
    const timeout = request.getUrl()?.getDefaulttimeout();
    const waitUntil = <'load' | 'domcontentloaded' | 'networkidle' | undefined>request.getWaituntil();
    if (url[0] === '/' && url.lastIndexOf('/') > 0) {
        const lastSlash = url.lastIndexOf('/');
        const re = new RegExp(url.substring(1, lastSlash), url.substring(lastSlash + 1));
        await page.waitForNavigation({ timeout, url: re, waitUntil: waitUntil });
    } else {
        await page.waitForNavigation({ timeout: timeout, url: url, waitUntil: waitUntil });
    }
    return emptyWithLog(`Navigated to: ${url}, location is: ${page.url()}`);
}
```

node/tsconfig.json
```diff
@@ -34,7 +34,7 @@
     // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
 
     /* Experimental Options */
-    "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
```
mkorpela, line 37: Did not get this tsconfig change.

mkorpela (approved):

mkorpela:
Thanks @vincenzo-gasparo for the PR!

mkorpela:
Failure in test is related to a flickering test case on windows with a timeout problem.

Issue #815, opened 2021-03-18
Add possiblity to set up screenshot directory


mnojek:
There is a really handy keyword in a similar library (😉) which is `Set Screenshot Directory`.
In BrowserLibrary it's possible to set the path to the screenshot explicitly in `path` argument when using `Take Screenshot` keyword but when the screenshot is taken for failed test the keyword is run through the default import parameter `run_on_failure` which does not support providing directory. I would like to be able to set the directory for all the screenshots and by default they should land in `${OUTPUT DIR}` in my opinion.
Would be nice to have similar keyword like `Set Screenshot Directory` here also.

aaltat:
Instead of mimicking SL, which may still be good idea,  would it be better to support defining arguments for the run on failure keyword?

mkorpela:
The merged PR #931 changes it so that registering failure keywords now allows arguments. It also enables absolute path for filename use in `Take Screenshot  filename`.
@mnojek will this be enough for your requirements?

mnojek:
Thanks @mkorpela for the changes. They are definitely needed but I still see the lack of one useful feature.
I would still like to have a way to set up a directory for screenshots globally, not only when taking screenshots explicitly.
By default the import parameter `run_on_failure` takes `Take Screenshot` keyword, but I cannot edit the place where I want them to be put. That I would like to control.

mkorpela:
There are many subdirectories under the main directory of Browser outputs.
Setting this main directory with `Set Browser Output Directory` would move screenshots, videos and other assets.

mkorpela:
We also discussed with @aaltat about possiblity to add arguments to `run_on_failure`. This might be another thing we will implement `run_on_failure=Take Screenshot \ yourpath/foo/bar.png`
The separator is still under discussion.

aaltat:
Arguments support for run on failure is in #1215

aaltat:
Is this still needed or can we close the issue?

mnojek:
I haven't used Browser library for a while so I don't want to decide. Someone else may check it and respond or you can close it if you want.

aaltat:
I guess we will close it, there has been enhancements on the area and need may not exist anymore. We can re-open this one or create new one.

PR #931, opened 2021-04-17
Fix/screenshot path

mkorpela:
There seems to be at least two problems:
1. Screenshot filename could not be an absolute path
2. Registering a screenshot as a fail keyword could not take args


mkorpela:
Relates to https://github.com/MarketSquare/robotframework-browser/issues/815

PR #1029, opened 2021-05-27
Add reject keyword for promises and get promise state

xylix:


xylix:
To actually support cancelling promises that have "started execution" we would probably need to move from concurrent.Futures to asyncio.futures Tasks. 

Documentation why we cannot cancel concurrent.Futures: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel

xylix:
We can re-open if we want to prioritize this in the future

PR #1034, opened 2021-05-28
Close playwright process on exit

### LINKED ISSUE #1033: Lingering node processes when running in Python
**Describe the bug**
When using the library in pure Python, without a Robot Framework execution, the library does not close its node processes.

All automatic clean-up is currently handled with Robot Framework listener methods, including killing the spawned subprocess on `close()`. This however does not happen inside Python, and this is not obvious to users (and the API isn't documented AFAIK).

**To Reproduce**

Naive implementation:
```python
import Browser

try:
    browser = Browser.Browser()
    browser.new_browser()
    # Do stuff
finally:
    browser.close_browser()
```

Working implementation:
```python
import Browser

try:
    browser = Browser.Browser()
    browser.new_browser()
    # Do stuff
finally:
    browser.playwright.close()
```

**Expected behavior**
I think the library should register an `atexit` function (or similar) to ensure the spawned node process is killed when the Python process exits.

 - OS: All
 - Browser: All
 - Version: 5.0.0

osrjv:
This PR fixes #1033 

Rundown of changes:
- Register Playwright `close` method to run on Python process exit
- Change `close` logic to always kill its subprocess, even if grpc throws an exception
- Change cached property access to not accidentally spawn a server on close
- Remove now superfluous listener method to not call `close` twice per process (which is ok but pointless)

Wasn't exactly sure where to put the atexit call, but in general it should be as late as possible because the registered methods are executed in LIFO order and we want our low-level stuff to still be available, but it also should always run even if something unexpected throws.

### DIFF
--- a/Browser/browser.py
+++ b/Browser/browser.py
@@ -712,12 +712,6 @@ class Browser(DynamicCore):
     def browser_output(self) -> Path:
         return Path(self.outputdir, "browser")
 
-    def _close(self):
-        try:
-            self.playwright.close()
-        except ConnectionError as e:
-            logger.trace(f"Browser library closing problem: {e}")
-
     def _start_suite(self, suite, result):
         if not self._suite_cleanup_done and self.browser_output.is_dir():
             self._suite_cleanup_done = True
--- a/Browser/playwright.py
+++ b/Browser/playwright.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import atexit
 import contextlib
 import os
 import time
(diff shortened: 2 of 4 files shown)


aaltat:
PR looks good from a mobile screen, but I need to take better look from bigger screen. 

You could think is the chance possible to test in https://github.com/MarketSquare/robotframework-browser/blob/master/utest/test_python_usage.py

mkorpela:
This is kind of special and easily accidentally broken. Test case would be nice.

osrjv:
The failed getter test doesn't seem to be related to this PR

aaltat:
Yes, there is some flakiness in our CI. Rerun test

aaltat:
Looks good, ready merge.

PR #1109, opened 2021-06-29
Properly implement `Browser.Browser().playwright.port` and add test

xylix:
- Working support for multiple Browser's using the same port of node side process also through library constructor
  - corresponding test
- Library method for overriding the node-side peer ID for browser processes. Useful for debug tooling or such which may need to access same browser/context/page between multiple separate processes.


aaltat:
I still do not see why this is done, could we discuss it bit? What is architecture and design idea?

mkorpela:
I agree with @aaltat . Supporting unknown external functionality is very bad idea. We need some concrete value out of a change.

PR #1207, opened 2021-08-04
Add node 16 to test-install CI-workflow

xylix:


xylix:
The problem here might be the unresolved promise rejections that node has been warning about for years now actually crashing the process.

xylix:
There is still some crash that is happening that needs fixing.

PR #1229, opened 2021-08-14
Set PLAYWRIGHT_BROWSERS_PATH to "0" only when not explicitly set by the user

### LINKED ISSUE #992: Setting browser path by PLAYWRIGHT_BROWSERS_PATH not working anymore
It is not possible anymore to change the location of playwright browser binaries with `PLAYWRIGHT_BROWSERS_PATH`. 

Before commit 391ddcf8efff24bf49fd71487c5414343e2899b0 it was possible. 

The referenced commit added the line 

```python
    os.environ["PLAYWRIGHT_BROWSERS_PATH"] = "0"
```

This **hardcodes** the binary path to the default: 

```python 
# lib\site-packages\Browser\wrapper\node_modules\playwright\lib\utils\registry.js

    const envDefined = utils_1.getFromENV('PLAYWRIGHT_BROWSERS_PATH');
    if (envDefined === '0') {
        result = path_1.default.join(__dirname, '..', '..', '.local-browsers'); # <<< always used
    }
    else if (envDefined) {
        result = envDefined; # <<< not possible anymore!
    }
```

I would expect Browser to behave as it did before this commit and to respect the variable `PLAYWRIGHT_BROWSERS_PATH` again to work as it is intended and documented by Playwright: https://playwright.dev/docs/installation/#managing-browser-binaries

ealap:
Fixes #992

Set `PLAYWRIGHT_BROWSERS_PATH` to `"0"` only when not explicitly set by the user

### DIFF
--- a/Browser/playwright.py
+++ b/Browser/playwright.py
@@ -93,7 +93,7 @@ class Playwright(LibraryComponent):
             os.environ["DEBUG"] = "pw:api"
         logger.info(f"Starting Browser process {playwright_script} using port {port}")
         self.port = port
-        os.environ["PLAYWRIGHT_BROWSERS_PATH"] = "0"
+        os.environ["PLAYWRIGHT_BROWSERS_PATH"] = os.getenv("PLAYWRIGHT_BROWSERS_PATH") or "0"
         return Popen(
             ["node", str(playwright_script), port],
             shell=False,


aaltat:
Basically this is OK, but I am thinking would it be possible to write a test for this?

PR #1276, opened 2021-09-03
Take no screenshots on failure

aaltat:


aaltat:
Lets think other ways to reduce the artifact size.

PR #1313, opened 2021-09-15
Terminate process if kill does not stop it

xylix:


xylix:
This might be useful if there is some actual case of the process "leaking" invalidly, but currently there's little data we have on this.

PR #1318, opened 2021-09-16
Start adding skip_if_exists option to New Browser

### LINKED ISSUE #1319: New Browser to skip if exists and switch

New Context by default either attaches to existing browser or creates a one with default options. For several cases default options are not good enough. For instance a headfull browser might be needed or a firefox.

add an option to skip and switch to `New Browser` that can find the right default browser for you.

xylix:
Closes #1319


xylix:
Needs a test for how list arguments for New Browser behave

xylix:
If we could do the equality check on Python side the code for it might be much simpler

Snooz82:
Closed due to new implementation

PR #1450, opened 2021-10-31
Wanted to use user data directory for persistency but got stuck..

paazmaya:
I need some input if this is the right approach to take, when I want to utilise persistent user data between sessions.


aaltat:
By a quick look, idea looks good to go. @xylix or @mkorpela  do you see something, because your experience is better on node side.

mkorpela:
Looks good. Tests fail

aaltat:
Closing this due inactivity.

paazmaya:
No worries, I will re-create when I have time to focus more.

PR #1511, opened 2021-11-17
Remove unnecessary comments, add 'Expect Failure' style test for too …

xylix:
…big upload


xylix:
There seems to be some weird extra errors on windows when uploading the 75MB test file. Need to fix before merging.

PR #1533, opened 2021-11-24
Fix wait for regex and predicate handling

rsandbach:
- Standardized urlOrPredicate handling
 - Corrected grammar/spelling in waitForRequest log message.
 - Added acceptance tests for URL, RegEx, and Predicate.

### DIFF
--- a/atest/test/03_Waiting/wait_for_http.robot
+++ b/atest/test/03_Waiting/wait_for_http.robot
@@ -19,15 +19,16 @@ Wait For Request async
 
 Wait For Request url
     Click    \#delayed_request
-    Wait For Request  matcher=${ROOT_URL}/api/get/json  timeout=1s
+    Wait For Request    matcher=${ROOT_URL}/api/get/json    timeout=1s
 
 Wait For Request regex
     Click    \#delayed_request
-    Wait For Request  matcher=\\/\\/local\\w+\\:\\d+\\/api  timeout=1s
-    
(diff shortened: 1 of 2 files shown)


aaltat:
Please run `inv lint` locally and push changes back to the PR.

rsandbach:
When I run `inv lint` , I get a message: `utest\test_waiters.py:13: error: Cannot assign to a method  [assignment]
Found 1 error in 1 file (checked 55 source files)` and it does not seem to run inv lint-robot. I ran it separately, and it modified all robot files. I committed only wait_for_http.robot.

aaltat:
`inv lint-robot` fixes only the robot file. 

But it is interesting that you get that error, because it works in CI and it worked yesterday in my Mac.

rsandbach:
> 

I did upgrade pip in my environment because there were some dependency issues with mypy. Let me recreate my venv and I'll see if that resolves the issue.

rsandbach:
I am actually unable to build on Windows without updating pip.

A possible explination for why CI passes is that it looks like CI does not run `inv lint`, it runs the mypy command separately and it does not include the utest/ directory.

**tasks.py**
`c.run("mypy --show-error-codes --config-file Browser/mypy.ini Browser/ utest/")`

**CI**
image

xylix:
I think running mypy for `utest` is something we need to fix on master. But I think the PR is working as expected now.

PR #1537, opened 2021-11-24
Update Generate Test File cmd to be OS dependent

rsandbach:
Windows doesn't support dd, added platform check and changed Windows to use fsutil.


aaltat:
@allcontributors please add @rsandbach for code

PR #1550, opened 2021-11-26
update dead link

Leemur89:
Link to playwright list of devices was dead, I've replaced it with the correct one (I already replaced in in a previous PR, but the links was appearing twice, here I'm updating the second dead link)


aaltat:
@allcontributors please add @Leemur89 for code.

PR #1551, opened 2021-11-26
Run `inv lint-python` in CI instead of custom mypy command

xylix:


aaltat:
Are you sure that inv lint fails the build if there is an error?

xylix:
Hmm. Need to confirm that. It does do that at least locally I think.

PR #1552, opened 2021-11-26
Initial attempt at generating releasenotes with tag difference instead of milestones

xylix:
TODO: Add some more structure, like rellu does (see https://github.com/MarketSquare/robotframework-browser/blob/main/docs/releasenotes/Browser-10.1.0.rst#id33 or https://github.com/robotframework/rellu/blob/master/rellu/releasenotes.py for examples)


xylix:
Sample of what this looks like: https://gist.github.com/xylix/0438a6219a21c9cce1b763929d2db24c

xylix:
Improved formatting, still need a way to get list of `acknowledgment` label commits / issues included in the release

PR #1573, opened 2021-11-29
Js docs example babel

mkorpela:


aaltat:
Should example folder go under docs/example?

PR #1577, opened 2021-11-30
Modify node/dynamic-test-app to accomodate esbuild, remove webpack config

xylix:


xylix:
Need to add some `path: path.resolve(__dirname, './dynamic-test-app/dist') },` magic or such to accomodate CI running the command from different working dir.

xylix:
Last missing piece seems to be that with the "unzipped demoapp" the test-installs seem to have problems with finding files that are in `static`

PR #1593, opened 2021-12-05
implemented Node side of Get Element States

Snooz82:
@mkorpela @aaltat 

May you have a look to the node side?


aaltat:
Hmmm, node side looks complex. Would be be cleaner to divide node side to methods? But I must say that I am not expert in node side and I would trust what Mikko has to say.

mkorpela:
Improvement ideas:
The binary enum thing is smart - too smart. Would use keys and values.
Try catch hell .. would try to change to something else like extracting methods or early exits.

PR #1597, opened 2021-12-05
Improve Select Options By documentation and examples

aaltat:


Browser/keywords/interaction.py
```diff
@@ -588,10 +591,19 @@ def select_options_by(
         If no values to select are passed will deselect options in element.
 
         Example:
-        | `Select Options By`    select[name=preferred_channel]    label    Direct mail
+        | ${selected} =    `Select Options By`    select[name=preferred_channel]    label    Direct mail
+        | List Should Contain Value    ${selected}    Direct mail
         | `Select Options By`    select[name=interests]    value    males    females    others
```
Snooz82, line 596: missing assignment to ${selected}

Browser/keywords/interaction.py
```diff
@@ -610,8 +622,8 @@ def select_options_by(
                     selector=selector, matcherJson=matchers, strict=self.strict_mode
                 )
             )
-            logger.debug(response.log)
-            return json.loads(response.json)
+        logger.debug(response.log)
+        return json.loads(response.json)
```
Snooz82, line 626: `response` variable is only on valid if the `with` is valid.
Why outside `with`

Snooz82 (changes requested):

PR #1621, opened 2021-12-15
Add custom_output_dir input param

### LINKED ISSUE #815: Add possiblity to set up screenshot directory

xylix:
Fixes #815


xylix:
Probably needs at least a test, and we need to define which artifacts should go here

Issue #1794, opened 2022-02-22
virtual_mouse.robot with keyword test that is always true


UliSei:
It is about virtual_mouse.robot, Test Case **Drag and Drop with coordinates**, lines 64 and 65.
The assertion is about testing whether the dragged object coordinates are the same as the coordinates of the destination object.
However the tests with given numbers for x and y ...
```
    Should Be True    ${x_diff} < 1 or ${x_diff} > -1
    Should Be True    ${y_diff} < 1 or ${y_diff} > -1
```
... are always true. 

**Suggestion:**
```
    Should Be True    -1 < ${x_diff} and ${x_diff} < 1
    Should Be True    -1 < ${y_diff} and ${y_diff} < 1
```

UliSei:
Test result with changed lines 64 and 65:
![grafik](https://user-images.githubusercontent.com/89480399/155298007-6980cce4-927a-42c6-8709-459416f01e46.png)

If there aren't any objections I'd create a PR.

aaltat:
Done in #1799

Issue #1716, opened 2022-01-26
Extension of keyword Drag And Drop By Coordinates: make release of the mouse button flexible


UliSei:
So far the keyword "Drag And Drop By Coordinates" does the following:

```
First it moves the mouse to the start-point,
then presses the left mouse button,
then moves to the end-point in specified number of steps,
then releases the mouse button.
```

I'd suggest to add a boolean parameter "release" which controls whether the mouse button at the end is released or not.

The default value would be True so that nothing would change for existing calls of the keyword.

The extended keyword used in sequence allows to drag the mouse several times to different coordinates before the mouse button is released.

This would allow a more realistic simulation of user interaction.

And it would allow to test GUI elements that react permanently on the position of the mouse like virtual joysticks.

Suggestion to Browser/keywords/interaction.py, not tested yet:

OLD:
```
def drag_and_drop_by_coordinates(
        self, from_x: float, from_y: float, to_x: float, to_y: float, steps: int = 1
    ):  ...
        self.mouse_button(MouseButtonAction.down, x=from_x, y=from_y)
        self.mouse_move(x=to_x, y=to_y, steps=steps)
        self.mouse_button(MouseButtonAction.up)
```

NEW:
```
def drag_and_drop_by_coordinates(
        self, from_x: float, from_y: float, to_x: float, to_y: float, steps: int = 1, release: bool = True
    ):  ...
        self.mouse_button(MouseButtonAction.down, x=from_x, y=from_y)
        self.mouse_move(x=to_x, y=to_y, steps=steps)
        if release:
            self.mouse_button(MouseButtonAction.up)
```

aaltat:
I do not see problems with such change.

UliSei:
Fine :-)
My first contributon was just about docu text, so I did not do any functional test with my forked version. This time I would have to test the change with a web application which is available in my docker environment. When I fork the library in GitHub and clone and change it locally - how could I test it? How do I have to handle the node dependencies? 
I offer to do the pull request but I would need some support.

aaltat:
You need to manually install Python and NodeJS, but rest of the development details are handled with scripts and invoke. Please see https://github.com/MarketSquare/robotframework-browser/blob/main/CONTRIBUTING.md#development-environment

We also have a inbuilt test application which provides different range of static and dynamic pages. Running tests is explained in here https://github.com/MarketSquare/robotframework-browser/blob/main/CONTRIBUTING.md#testing

[...]

PR #1799, opened 2022-02-23
Add drop=True arg for Drag And Drop By Coordinates (#1716 and #1794)

UliSei:
#1716 Add drop=True arg for Drag And Drop By Coordinates1716 drag and drop

issue [#1716](https://github.com/MarketSquare/robotframework-browser/issues/1716), keywords.resource:
ADD: new keyword **Assert Position**
**Assert Position** is used in Test Cases **Drag and Drop with coordinates** and **Drag And Drop**

virtual_mouse.robot:
ADD: Extend tests in **Drag and Drop with coordinates**
MOD: keyword **Drag And Drop**: switch from wrong assert to new keyword Assert Position.
-> solves also issue [#1794](https://github.com/MarketSquare/robotframework-browser/issues/1794)


UliSei:
I found some linting issues and wanted to correct them locally and commit.
upstream/main had changed in the meantime - not on files I worked on, I think. 
Wanted to pull the files from upstream/main to my local branch.
Unfortunately there is now a message:
grafik
How to continue?

aaltat:
Please run `inv lint` and push the changes.

UliSei:
I'll try and give a protocol how far I get.
Step 1: inv lint
grafik

grafik

Step 2: delete the ] and try again >>> inv lint
grafik

Step 3: test again >>> inv atest
excerpt:
grafik
-> errors probably due to Windows platform.
-> stop the process after ~20min.
grafik

Step 4: test the changed virtual_mouse only >>> inv atest -s virtual_mouse
grafik
-> 14 tests passed

UliSei:
The problem was the fast-forward option being active, [stackoverflow](https://stackoverflow.com/questions/13106179/fatal-not-possible-to-fast-forward-aborting#:~:text=This%20is%20because%20you%20have,at%20the%20time%20of%20pull.):
"Your branch is no longer directly based off of the branch you're trying to merge it into - e.g. another commit was added to the destination branch that isn't in your branch. Thus, you can't fast-forward into it (because fast-forward requires your branch to completely contain the destination branch)."

So I switched it off 
`git pull --no-ff`
and the pull from origin was possible :-)

aaltat:
The error in Mac looks random (We do have random errors not and then), but errors in Ubuntu looks like they are related. CI stores result as artifacts, are you able to download and review them?

UliSei:
Yes, I am going to use the artifacts once again. 
So far as I can see the coordinates which are determined in **Assert Position** are not correct with Ubuntu:
grafik
The draggable object as determined in the function seems not to be where it should be.
Eventually the wrong coordinates belong to a point from an earlier step.

When the "Should Be True" statements were not corrected they would give "True". 
Maybe there were hidden errors in the past.

It's not obvious where the different behaviour comes from.
I'll have to try with more simple test cases.
Unfortunately I have no local Ubuntu environment so that I'll have to send some more commits which will probably break again the pipeline.

UliSei:
The bad thing is that there is now also an error in WIN env.
The assertion of the position is OK but the next Drag ends with unknown error:

grafik
It must be assured that the **Assert Position** test function does not have any impact on later keywords.

UliSei:
Interesting: I call "Drag And Drop By Coordinates" several times without the new drop argument and it fails with a command that was OK 6 keywords before:
grafik

The message "No coordinates where set. Action appears at current position." ...
grafik

... comes from the action Mouse Up I assume?
grafik

No idea how to detect the unknown error which occurred instead.
Never debugged RF so far and I try to avoid it ;-)

UliSei:
I switched back to the former version of the keyword, without drop argument.
And I tested with a series of more than 10 Drag & Drops to 5 different points P1 - P5.
Test Case _Drag and Drop with coordinates_ fails even in my Windows environment when the order is from P1 to P2, P3, P4, P5, P1, P2, **P3**.
It also fails with changed order
from P1 to P2, P3, P4, P5, P1, P3, **P2**  and also
from P1 to P5, P4, P3, P2, P1, P5, **P4**
Then I separated the _Drag And Drop By Coordinates_ keyword by _Sleep_ commands.
This time the unknown error message came earlier within the 5. move:
grafik

The source for the message is playwright.py line 155.
grafik

From my point of view we face an issue that pops up within the playwright context manager. 
It seems to be an issue with the old function that was revealed by an increased number of tests.

aaltat:
Most likely better error can be found form the Playwright debug logs.

UliSei:
Just had a look to playwright-log.txt.
It contains things like a nonexisting locator (which could be OK because there are just coordinates for the move, no locator?).
But there are also timeouts.
Please find attached a longer part of the log. The complete log would be about 7MB.

```
2022-02-25T11:55:09.641Z pw:api => locator.click started
2022-02-25T11:55:09.642Z pw:api waiting for selector ".nonexisting2"
2022-02-25T11:55:09.652Z pw:api   "domcontentloaded" event fired
2022-02-25T11:55:10.138Z pw:api   "networkidle" event fired
2022-02-25T11:55:12.780Z pw:api <= locator.click failed
{"level":30,"time":"2022-02-25T11:55:12.781Z","pid":1942,"hostname":"Mac-1645788844395.local","msg":"Error of node method  click"}
================= Original suppressed error =================
locator.click: Timeout 3000ms exceeded.
=========================== logs ===========================
waiting for selector ".nonexisting2"
============================================================
    at click (/Users/runner/work/robotframework-browser/robotframework-browser/Browser/wrapper/index.js:6739:18)
    at async PlaywrightServer.click (/Users/runner/work/robotframework-browser/robotframework-browser/Browser/wrapper/index.js:7507:28) {
  name: 'TimeoutError'
}
=============================================================
{"level":30,"time":"2022-02-25T11:55:12.798Z","pid":1942,"hostname":"Mac-1645788844395.local","msg":"Start of node method takeScreenshot"}
2022-02-25T11:55:12.798Z pw:api => page.screenshot started
2022-02-25T11:55:13.090Z pw:api <= page.screenshot succeeded
``
[playwright-log-excerpt.txt](https://github.com/MarketSquare/robotframework-browser/files/8141251/playwright-log-excerpt.txt)
`

**Comment by UliSei** [comment]
The failed pipeline for Mac OS is not because of my development.
virtual_mouse.robot has PASSed:

grafik

**Comment by UliSei** [comment]
The tests seem to be a little bit flaky.
This time the Mac OS and Ubuntu tests passed, but there were timeouts with win32-rf-4.1.3-py-3.9.10-node-v12.22.10 and win32-rf-4.1.3-py-3.8.10-node-v12.22.10, see "Test.01 Browser Management.Switching Browsers" for both:

grafik

**Comment by UliSei** [comment]
It's hard to get all theses jobs passed!

grafik

**Comment by UliSei** [comment]
Two times the installation step failed because of a timeout in Test.01 Browser Management.Switching Browsers
How do you get along with it?

grafik

**Comment by aaltat** [comment]
It might not be related, GitHub actions have been unstable for the whole week and I have seen various errors in jobs.

**Comment by UliSei** [comment]
Tried today to merge the sources as before.
There is progress - just one job in step 2 that failed.
Maybe all jobs are passing next time.

grafik

**Comment by UliSei** [comment]
Success!
Please decide whether or how to reset the timeout parameter. Its original value was 10s.
grafik

File: atest/test/02_Content_Keywords/virtual_mouse.robot [hunk 1]
```diff
@@ -53,16 +54,73 @@ Drag And Drop With Strict
...
+    ${x3} =    Evaluate    ${x2} + 0.06 * ${width}
+    ${y3} =    Evaluate    ${y2} + 1.2 * ${height}
+    ${x4} =    Evaluate    ${x3} + 0.08 * ${width}
+    ${y4} =    Set Variable    ${y3}
+    ${x5} =    Set Variable    ${x4}
+    ${y5} =    Set Variable    ${height}
+    ${steps} =    Set Variable    200
+    ${time} =    Set Variable    1
+
+    # Temporary deactivated. The very next call of "Drag And Drop By Coordinates" would make MacOS fail.
```
aaltat, line 87: These should not be commented out. Do we still need these?

aaltat (changes requested):

atest/test/02_Content_Keywords/virtual_mouse.robot
UliSei: We can divide it in an own issue. 
There is a bug when ones repeats the call of the keyword certain times.
Mac OS is the most sensitive OS in that case but the failures also occur with WIN and Ubuntu some calls later.
I'll do some more tests anyway when I use the new keyword argument with my application.
Maybe that I hereby get more insights when the issue occurs or not.
So I'd delete the commented part for now for that we have clean code.

UliSei: Need to be patient with this PR :-)
Build processes and test install were successful but there was one failure dealing with uploading artifacts,
see test-install (windows-latest, 3.9, 16.x) 

For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging

grafik

If I should take further steps please let me know.

UliSei:
So all the checks were ok today  :-)

atest/test/02_Content_Keywords/virtual_mouse.robot
aaltat: 1) Tests
Generally I feel that commenting code/test out is not good idea. Either code/tests are valid and is run or it is not needed and should be deleted. It can be always found from the history if there is need to dig it out. Many cases there is not
If these are valid test, then we should run them. We have test which are not successful in some platforms and those test can be skipped by using `no-windows-support` or `no-mac-support` tags. 
I think we should enable the tesst, but skip them in Mac and Win OS.

2) Failures
GitHub actions have had some regression in performance and some steps in CI just fail without a good reason. Currently actions seems to be more stable, but if you see error in steps which is running tests, we can ignore the error.

PR #1801, opened 2022-02-24
Make inv atest-robot execution to use --console dotted

aaltat:
For better readability in CI.


aaltat:
This does not work well, lets not do it.

Issue #1685, opened 2022-01-17
Python script promise_to always got error "No library '<Browser.browser.Browser object at 0x...>' found."


Massukio:
**Describe the bug**
Python script promise_to always got error "No library '<Browser.browser.Browser object at 0x...>' found."

**To Reproduce**
test.py
```
from robot.api.deco import keyword
from Browser import Browser

@keyword
def test_keyword():
    lib = Browser()
    lib.new_browser(headless=False)
    lib.new_page(url='http://www.google.com')
    promise = lib.promise_to('Wait For Response')
    body = lib.wait_for(promise)
```
test.robot
```
*** Settings ***
Library  test.py

*** Test Cases ***
Test
    Test Keyword
```
test.robot result
```
==============================================================================
Test
==============================================================================
Test                                                                  | FAIL |
No library '<Browser.browser.Browser object at 0x000001B127FC01F0>' found.    
------------------------------------------------------------------------------
Test                                                                  | FAIL |
1 test, 0 passed, 1 failed
==============================================================================
```

**Expected behavior**
Run keyword succeed and the promise_to works properly

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: chrome
 - Version: robotframework-browser==11.1.1

**Additional context**
Use the "Promise To" in robot file, the test can be executed and pass properly

test2.robot
```
*** Settings ***
Library  Browser

*** Test Cases ***
Test
    New Browser
    New Page    url=http://www.google.com
    ${promise}=  Promise To    Wait For Response
    ${body}=  Wait For  ${promise}
```
test2.robot result
```
==============================================================================
Test2
==============================================================================
Test                                                                  | PASS |
------------------------------------------------------------------------------
Test2                                                                 | PASS |
1 test, 1 passed, 0 failed
==============================================================================
```

aaltat:
@allcontributors please add @Massukio for bugs

luchoagomezt:
> This is true and I can reproduce the problem. The problem lies in [here](https://github.com/MarketSquare/robotframework-browser/blob/main/Browser/keywords/promises.py#L50), library expect that it is imported in RF side and finds itself. Most likely you can workaround the problem by importing the Browser library in RF side and using this `lib = BuiltIn().get_library_instance("Browser")`.
> 
> But your way should be made possible and therefore I am considering this a bug.

@aaltat hello, I would like to help fixing this issue.  Are the instructions in CONTRIBUTING.md all that I need to follow?

[...]

PR #1806, opened 2022-02-25
Fixing issue #1685, "No library '<Browser.browser.Browser object at 0x...>' found."

luchoagomezt:
@aaltat Please review the code here, thanks


Browser/keywords/promises.py
```diff
@@ -47,6 +48,25 @@ def promise_to(self, kw: str, *args) -> Future:
...
+            "wait for navigation": self.library.wait_for_navigation,
+            "wait for request": self.library.wait_for_request,
+            "wait for response": self.library.wait_for_response,
+        }
+        if EXECUTION_CONTEXTS.current:
+            promise = self._robot_promise_to(kw, args)
+        elif kw.strip().lower() in library_method:
+            promise = self._executor.submit(library_method[kw.strip().lower()], *args)
+
+        self.unresolved_promises.add(promise)
```
aaltat, line 64: There is a bug: `promise` variable does not exist in all cases.

Browser/keywords/promises.py
```diff
@@ -47,6 +48,25 @@ def promise_to(self, kw: str, *args) -> Future:
...
         """
+        library_method = {
+            "wait for": self.library.wait_for,
+            "wait for alert": self.library.wait_for_alert,
+            "wait for function": self.library.wait_for_function,
+            "wait for navigation": self.library.wait_for_navigation,
+            "wait for request": self.library.wait_for_request,
+            "wait for response": self.library.wait_for_response,
+        }
+        if EXECUTION_CONTEXTS.current:
```
aaltat, line 59: This `if/else` logic does not solve the problem when library is used in the way described https://github.com/MarketSquare/robotframework-browser/issues/1685#issue-1105409364 Instead trying to solve it like this, it would better look https://github.com/robotframework/PythonLibCore/blob/main/src/robotlibcore.py direction and see how it can be solved in unified way when using library purely from Python and with Robot Framework. 

Similar problem is already solved when registering keyword to run on failure. Looking in this direction: https://github.com/MarketSquare/robotframework-browser/blob/main/Browser/browser.py#L741 and making it more generic would perhaps be the best option.

utest/test_python_usage.py
```diff
@@ -76,3 +76,18 @@ def test_promise_handling(browser, application_server):
     browser.promise_to_upload_file(file.resolve())
     browser.click("#file_chooser")
     assert browser.get_text("#upload_result") == "test_python_usage.py"
+
+
+def test_wait_for_promise_to(browser):
```
aaltat, line 81: There should be also acceptance test, you can copy paste the example from the issue https://github.com/MarketSquare/robotframework-browser/issues/1685#issue-1105409364

aaltat (changes requested):
The problem described with the `if/else` is the most important problem to solve. Once that problem is solved, we can look are the rest of the comments/code important to solve.

Browser/keywords/promises.py
```diff
@@ -28,6 +28,7 @@
 
 class Promises(LibraryComponent):
     def __init__(self, library):
+        self.library = library
```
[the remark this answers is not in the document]
luchoagomezt: I removed line 31

Browser/keywords/promises.py
```diff
@@ -47,6 +48,25 @@ def promise_to(self, kw: str, *args) -> Future:
         | `Click`           \\#delayed_request
         | ${body}=        `Wait For`              ${promise}
         """
+        library_method = {
```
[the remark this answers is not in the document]
luchoagomezt: Please have a look at the update, in particular to  def resolve_arguments(*args) and promises.py line 57

luchoagomezt:
> The problem described with the `if/else` is the most important problem to solve. Once that problem is solved, we can look are the rest of the comments/code important to solve.

I removed the if/else block, both python and robot use the same code

aaltat:
There is linting error, please run `inv lint` command and commit to fix the problem.

luchoagomezt:
Aceptance test cases that use **Promise To** and **Wait For** must still pass, for instance "Shows a cat and a dog"

luchoagomezt:
> The problem described with the `if/else` is the most important problem to solve. Once that problem is solved, we can look are the rest of the comments/code important to solve.

Code contains one block for both the robot and the oython interface

luchoagomezt:
Lint is failing due to errors in atest/test/03_Waiting/wait_for_http.robot; I did not change that file

aaltat:
That should have been fixed in upstream, please rebase. It was caused by a  Tidy update.

luchoagomezt:
> That should have been fixed in upstream, please rebase. It was caused by a Tidy update.

Thanks, I rebased my repo; all the jobs are green

aaltat:
@allcontributors please add @luchoagomezt for code

aaltat (approved):

PR #1825, opened 2022-03-02
Allow config options for presenter mode

### LINKED ISSUE #1816: Should be able to change color and timeout for "enable_presenter_mode" argument in the "Library"
Yes, the default color for **"enable_presenter_mode"** argument in the "Browser Library" is blue and its getting mixed with html elements in our UI which are of same color 

I would like to be able to change the default color, **change the border style as well (solid border for example)** and time out for "enable_presenter_mode" argument in the "Browser Library"

I am using "Highlight Elements" keyword instead

As you could see in the screenshot, Yes button is in blue and cant see it highlighted when used "enable_presenter_mode"

image

aaltat:
Fixes #1816

### DIFF
--- a/Browser/base/librarycomponent.py
+++ b/Browser/base/librarycomponent.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 import os
 from concurrent.futures._base import Future
-from copy import deepcopy
+from copy import copy, deepcopy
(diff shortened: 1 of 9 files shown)


aaltat:
Missing docs at least.

PR #2009, opened 2022-05-03
Fixing issue 1980, action values in lower and upper case

luchoagomezt:
Both
    ${promise} =    Promise To    Wait For Alert    action=ACCEPT
and 
    ${promise} =    Promise To    Wait For Alert    action=accept

are valid


aaltat:
Hmm, this branch can not be merged to main code line because there are conflicts. Could you resolve the problem and write a test for the problem, so that similar problem does not happen again.

luchoagomezt:
> Hmm, this branch can not be merged to main code line because there are conflicts. Could you resolve the problem and write a test for the problem, so that similar problem does not happen again.

I fixed the conflict and added two test cases: a) action=ACCEPT and b) action=DISMISS

aaltat:
@allcontributors please add @luchoagomezt for code.

Issue #1961, opened 2022-04-14
Drag And Drop By Coordinates: drop arg does not work for GUI element "virtual joystick"


UliSei:
[Drag And Drop By Coordinates](https://marketsquare.github.io/robotframework-browser/Browser.html#Drag%20And%20Drop%20By%20Coordinates) tested with chromium and webkit

## Short overview:
it makes a difference to test the keyword
•	with a draggable element which stays where it was dropped
•	or with a virtual joystick which tends to return to its default position when it is dropped
Suggested solution see below

## Description of the system under test and a speculation why it behaves unexpectedly
There was a new **drop** argument introduced with  v12.2.0 which defaults to True. It was tested with the draggable GUI element and obviously worked fine. However the keyword does not meet the expectations when the GUI element moves by itself to a default position when it is dropped. Testing with a virtual joystick showed, that a sequence of _Drag And Drop By Coordinates_ statements with drop=False had the same behaviour as with drop=True. It looks as if the Browser Lib keyword _Drag And Drop By Coordinates_ is not yet designed to continue a previous call of _Drag And Drop By Coordinates_ with drop=False. The call of _Drag And Drop By Coordinates_ seems to expect and initialize a dropped mouse. 

For moving the draggable element it makes no difference whether drop is set to True or False. However when a virtual joystick is moved outside its center and the next _Drag And Drop By Coordinates_ is called, the test shows:
- the joystick first returns to its origin, even when drop was set to False. 
- In consequence the next _Drag And Drop By Coordinates_ is executed with a mouse that is not centered on the joystick. 
- When the first movement was long enough the joystick even does not move any more. This happens when the from coordinates don’t hit the joystick in its default position. 

## Test case with expected behaviour
Drag a GUI element, the lightened disk, from its origin O to point D passing the points A, B and C. 
```
Drag And Drop By Coordinates    ${Ox}    ${Oy}    ${Ax}    ${Ay}    10    False
Drag And Drop By Coordinates    ${Ax}    ${Ay}    ${Bx}    ${By}    10    False
Drag And Drop By Coordinates    ${Bx}    ${By}    ${Cx}    ${Cy}    10    False
Drag And Drop By Coordinates    ${Cx}    ${Cy}    ${Dx}    ${Dy}    10    False
Drag And Drop By Coordinates    ${Dx}    ${Dy}    ${Ox}    ${Oy}    10    True
```
[...]

PR #2030, opened 2022-05-17
issues/1961 solved

UliSei:
#1961 keyword Drag And Drop By Coordinates:
Removed drop parameter because it did not fulfill the expectations.

### DIFF
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -914,14 +914,13 @@ class Interaction(LibraryComponent):
         to_x: float,
         to_y: float,
         steps: int = 1,
-        drop: bool = True,
     ):
         """Executes a Drag&Drop operation from a coordinate to another coordinate.
 
         First it moves the mouse to the start-point,
         then presses the left mouse button,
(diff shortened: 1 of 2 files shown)


aaltat:
We can't do this in minor release, because it's backwards incompatible change. If the change is generally useful, I would leave the code as is, because it's a minor thing from maintenance point of view. Also getting rid off the argument can cause problems for other users.

UliSei:
Sure, it's your decision whether or when to implement the PR.
In my opinion the argument offers no benefit. The only use case (where it was designed for, was moving a draggable object that returns to its original position when the mouse is dropped. The drop=False argument was meant as a feature to move the draggable object in a sequence to arbitrary different positions before it returns to its origin.
Unfortunately this did not work. So if we leave the argument as it is we would have to make it work. Otherwise the argument is useless. That's why I bet that nobody uses the argument so far.

Snooz82 (approved):

aaltat:
I did not know that it doesn't work, well in that case let's merge it. Also next release will be 13.0.

aaltat:
Fixes #2042

Issue #2077, opened 2022-06-01
Select Option By causes super long time due to iteration over all options.


Snooz82:
**Describe the bug**
When a <select> element has really many options this keyword can easily take 20 Seconds.

Due to the iteration over all options.
But there is js function `getSelectedOptions` that is much much faster.

Example Page

https://www.provinzial.de/west/versicherung/kfz-versicherung/rollerund-mopedversicherung/rollerundmoped-abschluss.html

On Page 3 `Hersteller Ihres Fahrzeuges` ( `//*[@name="hersteller"]//option` ) has around 460 options.

PR #2078, opened 2022-06-01
Improve Speed of Select Option setter and getter #2077

Snooz82:
`Get Select Options` , `Get Selected Options` and `Select Options By` have been refactored and massively speeded up!

The real issue only happened with select lists with many options like 400.
On a productive Angular Page, it took more than 20 seconds to select an element.

Fixed now.


Snooz82:
In my opinion it is ready and can be merged and maybe also released as BugFix

Browser/keywords/getters.py
```diff
@@ -558,15 +558,14 @@ def get_selected_options(
...
-        if option_attribute is SelectAttribute.value:
-            selected = [sel.value for sel in response.entry if sel.selected]
-        elif option_attribute is SelectAttribute.label:
-            selected = [sel.label for sel in response.entry if sel.selected]
-        elif option_attribute is SelectAttribute.index:
-            selected = [
-                index for index, sel in enumerate(response.entry) if sel.selected
-            ]
+
+        if option_attribute is SelectAttribute.index:
```
aaltat, line 562: Instead of `is` should we use `==` in this case?

aaltat (commented):
No idea about node side, ask other team members.

Browser/keywords/getters.py
Snooz82: https://docs.python.org/3/library/enum.html#comparisons

as far as i understood it does not make any difference.
And in that case i do like the is more than the ==.
Because that shows that this is really not comparing a value of a variable to be equal but expecting a specific identity of enum member.

Issue #2081, opened 2022-06-02
Wait For Navigation fails: 'str' object has no attribute 'name'


ealap:
**Describe the bug**
Promise to wait for navigation fails due to library error:
```
'str' object has no attribute 'name'
```

**To Reproduce**
```
Click "${locator}" And Wait For Navigation To "${target}" Page Until "${event}"
    ${target_url} =    Resolve Full Url    ${target}
    ${page_navigation} =    Promise To    Wait For Navigation    url=${target_url}    wait_until=${event}
    Click    ${selector}
    Wait For    ${page_navigation}
```
`wait_until` has value `load` during test.

**Expected behavior**
Promise to wait for navigation should be successful.

**Traceback**
```
Traceback (most recent call last):
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/playwright.py", line 149, in grpc_channel
    yield playwright_pb2_grpc.PlaywrightStub(self._channel)
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/keywords/network.py", line 245, in wait_for_navigation
    waitUntil=wait_until.name,
AttributeError: 'str' object has no attribute 'name'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/browser.py", line 957, in run_keyword
    raise e
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/browser.py", line 947, in run_keyword
    return DynamicCore.run_keyword(self, name, args, kwargs)
  File "/home/robot/testenv/lib/python3.10/site-packages/robotlibcore.py", line 94, in run_keyword
    return self.keywords[name](*args, **(kwargs or {}))
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/keywords/promises.py", line 166, in wait_for
    return promises[0].result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 439, in result
    return self.__get_result()
  File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 391, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/robot/testenv/lib/python3.10/site-packages/Browser/keywords/network.py", line 238, in wait_for_navigation
    with self.playwright.grpc_channel() as stub:
[...]

PR #2100, opened 2022-06-10
Fix for Wait For Navigation fails: 'str' object has no attribute 'name' #2081

luchoagomezt:


Browser/keywords/promises.py
```diff
@@ -65,18 +65,22 @@ def promise_to(self, kw: str, *args) -> Future:
...
         positional: List[Any] = []
-        named = {}
+        named: Dict[str, Any] = {}
         logger.debug(f"*args {args}")
 
         keyword_arguments = [
             argument[0] if isinstance(argument, tuple) else argument
             for argument in self.library.get_keyword_arguments(kw)
         ]
+        logger.debug(f"keyword_arguments {keyword_arguments}")
```
aaltat, line 75: This should be removed.

Browser/keywords/promises.py
```diff
@@ -65,18 +65,22 @@ def promise_to(self, kw: str, *args) -> Future:
...
         for arg in args:
+            logger.debug(f"arg {arg}")
```
aaltat, line 77: This should be removed.

aaltat (changes requested):
Code looks OK, only debug logging statements should be removed. But the commit history is not good and it should contains only the commit(s) that contains the changes.

PR #2141, opened 2022-06-27
altered examples of Close Page

leeuwe:
Examples of Close Page now better reflects contexts and browser are not closed when using close page.


aaltat:
Fixes #2142

PR #2319, opened 2022-09-21
Added Scope feature to timeout, retry_assertion and strice_mode

### LINKED ISSUE #2327: Switch Page fails when context and browser id are given as well.
context id and Browser id could not be set.
Only CURRENT and ALL did work correctly

Snooz82:
With this feature a scope is added to the following keywords:

- [x] [`Set Browser Timeout`](https://browser-libdoc.surge.sh/#Set%20Browser%20Timeout)
- [x] [`Set Strict Mode`](https://browser-libdoc.surge.sh/#Set%20Strict%20Mode)
- [x] [`Set Retry Assertions For`](https://browser-libdoc.surge.sh/#Set%20Retry%20Assertions%20For)

With that scope argument one can set the "live time" of that setting.
Available Scopes are: `Global`, `Suite` and `Test`/`Task`

Live Times:
- A `Global` scope will live forever until it is overwritten by another `Global` scope.
- A `Suite` scope will locally override the `Global` scope and live until the end of the Suite within it is set, or if it is overwritten by a later setting with `Global` or same scope. Children suite does inherit the setting from the parent suite but also may have its own local `Suite` setting that then will be inherited to its children suites.
- A `Test` or `Task` scope will be inherited from its parent suite but when set, lives until the end of that particular test or task.

A new set higher order Scope will always remove the lower order scope which may be in charge.
So the setting of a `Suite` scope from a test, will set that scope to the robot file suite where that test is and removes the `Test` scope that has been in place.

### Bugfixes:
- Fixes #2327 (added tests)


Browser/browser.py
```diff
@@ -957,14 +965,58 @@ def _start_suite(self, name, attrs):
...
-    def _start_test(self, name, attr):
+    def _start_test(self, name, attrs):
+        self.timeout_stack.start(attrs["id"], Scope.Test)
         if self._auto_closing_level == AutoClosingLevel.TEST:
             try:
                 self._execution_stack.append(self.get_browser_catalog())
             except ConnectionError as e:
                 logger.debug(f"Browser._start_test connection problem: {e}")
 
+    def _start_keyword(self, name, attrs):
```
Snooz82, line 976: This code is just moved and unaltered from the lower part of that file, so that it is in a better place between start  and end test

Browser/keywords/getters.py
```diff
@@ -509,6 +509,7 @@ def get_selected_options(
         option_attribute: SelectAttribute = SelectAttribute.label,
         assertion_operator: Optional[AssertionOperator] = None,
         *assertion_expected,
```
Snooz82, line 511: The message argument was just missing...

mkorpela (approved):

PR #2367, opened 2022-10-10
connect to 127.0.0.1

slavaGanzin:
Not all servers has alias from localhost to 127.0.0.1.

I was getting mysterios "Could not connect to the playwright process at port 9000.", which indeed was:

DNS resolution failed for localhost:9000: C-ares status is not ARES_SUCCESS qtype=A name=localhost is_balancer=0: Domain name not found

hidden inside this [try](https://github.com/MarketSquare/robotframework-browser/compare/main...slavaGanzin:robotframework-browser:switch-from-localhost-to-127001?expand=1#diff-b6c4f37c431ecd50f11a5d0149e4f753f91cf52dd6fa2c38e31227282be580a6R115)


xylix:
Seems better! Thanks for the fix, will merge once the CI run passes.

xylix:
There seems to be another `localhost` reference in our code here https://github.com/MarketSquare/robotframework-browser/blob/90ebd8833d54a1e1ad7640ac6734f51d6f1a20a8/node/playwright-wrapper/index.ts#L31 . Did this work on your machine? I wonder if node does some custom routing to make that work on every machine. But we could change this as well to use `127.0.0.1`

slavaGanzin:
Now it binds to 127.0.0.1

xylix:
@allcontributors Add @slavaGanzin for bug. Add @slavaGanzin for code.

xylix:
@allcontributors Add @slavaGanzin for bug. Add @slavaGanzin for code.

slavaGanzin:
@xylix I don't care about it. But I was added to ideas, and not to code. So it seems bot is a bit buggy

PR #2613, opened 2023-01-29
Get Style with Pseudo Elements

Snooz82:
Get Style is now able to return any style property, not only those included in the dictionary.
Also Pseudo Elements are supported.

Also added:
- Added Scope to `Show Keyword Banner`

Also fixed:
- Fixed Logging issue with `Type/Fill Secret`
- Fixed Docs of Scope
- Fixed 'none' string issue with Assertions

### DIFF
--- a/atest/test/02_Content_Keywords/assertions.robot
+++ b/atest/test/02_Content_Keywords/assertions.robot
@@ -23,7 +23,7 @@ Get Attribute Names Does Not Support Formatter:
     Get Attribute Names    id=two    ==    id
     [Teardown]    Formatter TearDown
 
-Get Attribute Names Does Not Support Formatter:
+Get Attribute Names Does Not Support Formatter2:
     [Documentation]
     ...    LOG 1:*    DEBUG    GLOB:    Assertion polling statistics:*
     [Setup]    Go To    ${SPACES_URL}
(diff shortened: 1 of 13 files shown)


atest/test/02_Content_Keywords/assertions.robot
```diff
@@ -23,7 +23,7 @@ Get Attribute Names Does Not Support Formatter:
     Get Attribute Names    id=two    ==    id
     [Teardown]    Formatter TearDown
 
-Get Attribute Names Does Not Support Formatter:
+Get Attribute Names Does Not Support Formatter2:
```
aaltat, line 26: Why `2` at the end?

atest/test/02_Content_Keywords/assertions.robot
```diff
@@ -45,15 +45,51 @@ InEqual
 
 Greater Than
     Get Style    //*[@id="progress"]    width    >    100
```
aaltat, line 47: You could change this too?

aaltat (approved):

atest/test/02_Content_Keywords/assertions.robot
Snooz82: Because it had duplicate names. 
just cosmetics

atest/test/02_Content_Keywords/assertions.robot
Snooz82: It was three times xpath.

now it is xpath, css and id

thats why i did not change all

PR #2616, opened 2023-01-30
[Fixed #2570] Get code completion for LibraryComponent.library

### LINKED ISSUE #2570: [Bug] Get code completion for LibraryComponent.library does not work correctly
this is just a notice from RoboCon

If I rename__init__.pyi to browser.pyi it works for VSCode.

Snooz82:
Signed-off-by: René <snooz@posteo.de>

### DIFF
--- a/Browser/entry.py
+++ b/Browser/entry.py
@@ -206,11 +206,11 @@ def show_versions():
     version_file = CURRENT_FOLDER / "version.py"
     version_text = version_file.read_text()
     match = re.search(r"\"\d+\.\d+.\d+\"", version_text)
-    browser_lib_version = match.group(0)
+    browser_lib_version = match.group(0) if match else "unknown"
     package_json = INSTALLATION_DIR / "package.json"
     package_json_data = json.loads(package_json.read_text())
     match = re.search(r"\d+\.\d+\.\d+", package_json_data["dependencies"]["playwright"])
-    pw_version = match.group(0)
+    pw_version = match.group(0) if match else "unknown"
     logging.info(
         f'Installed Browser library version is: {browser_lib_version} with RF "{rf_version.VERSION}"'
     )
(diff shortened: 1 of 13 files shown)


Browser/entry.py
```diff
@@ -206,11 +206,11 @@ def show_versions():
     version_file = CURRENT_FOLDER / "version.py"
     version_text = version_file.read_text()
     match = re.search(r"\"\d+\.\d+.\d+\"", version_text)
-    browser_lib_version = match.group(0)
+    browser_lib_version = match.group(0) if match else "unknown"
     package_json = INSTALLATION_DIR / "package.json"
     package_json_data = json.loads(package_json.read_text())
     match = re.search(r"\d+\.\d+\.\d+", package_json_data["dependencies"]["playwright"])
-    pw_version = match.group(0)
+    pw_version = match.group(0) if match else "unknown"
```
aaltat, line 213: I would just raise an exception, it gets noticed better in that case

Browser/entry.py
```diff
@@ -206,11 +206,11 @@ def show_versions():
     version_file = CURRENT_FOLDER / "version.py"
     version_text = version_file.read_text()
     match = re.search(r"\"\d+\.\d+.\d+\"", version_text)
-    browser_lib_version = match.group(0)
+    browser_lib_version = match.group(0) if match else "unknown"
```
aaltat, line 209: I would just raise an exception, it gets noticed better in that cas

atest/test/09_Plugins/plugin.robot
```diff
@@ -55,7 +55,7 @@ Pluging Keyword Example Location
     Should Be Equal    ${location.protocol}    http:
     Should Be Equal    ${location.href}    ${FORM_URL}
 
-# TODO  ADD THESE TESTS WHEN NEW PLC IS RELEASED
+# TODO    ADD THESE TESTS WHEN NEW PLC IS RELEASED
```
aaltat, line 58: Is the comment correct and test could be returned?

Browser/entry.py
Snooz82: yea but it is just an information to the user.
Therefore the user would get scared by that exception.

PR #2715, opened 2023-03-24
Topic/ruff

### LINKED ISSUE #2717: Get Cookies returns a datetime object without time zone information
Get Cookies returns no time zone information (tzinfo) in the datetime object.

This does only affect the dict return type.
It should be fixed so, that the time can be calculated relative to your current time zone

Snooz82:
Switch to ruff

the one pytest is failing due to added timezone; which imho is important but one of you may decide if we want to change that.

Fixes #2717 as well


tasks.py
```diff
@@ -593,8 +593,9 @@ def lint_python(c):
             "mypy --exclude .venv --show-error-codes --config-file Browser/mypy.ini Browser/"
         )
         c.run("black --config Browser/pyproject.toml tasks.py Browser/")
-        c.run("flake8 --config Browser/.flake8 Browser/ utest/")
-        c.run("isort Browser/")
+        # c.run("flake8 --config Browser/.flake8 Browser/ utest/")
```
aaltat, line 596: Remove these.

.github/workflows/on-push.yml
```diff
@@ -157,9 +157,9 @@ jobs:
       if: matrix.shard == 1
       run: |
         inv lint-node
-    - name: Check with flake8
+    - name: Check with ruff
       run: |
-        flake8 --config Browser/.flake8 Browser
+        ruff Browser
```
aaltat, line 162: Utest too perhaps?

PR #2888, opened 2023-06-24
Supress locator.click: Target closed error

### LINKED ISSUE #2089: Can not close context if web browser tab was auto-closed
**Describe the bug**
Given: I have two contexts open. Context 1 and context 2.
When: I open extra tab in context 2
Then: I do some changes and extra tab is auto closed after saving some changes
And Then: I can't properly close the context 2 by using `Close Context`
And Then: I receive error `Error: page.bringToFront: Target page, context or browser has been closed.`
And Then: Context 1 is properly closed

**To Reproduce**
1. Run Browser
2. Prepare Context 1
3. Prepare Context 2
4. Open extra tab in context 2 and continue to its auto-close after e.g. saving changes
5. In Teardown close context 2
6. Switch to context 1
7. Close context 1

**Expected behavior**
Context 2 should be closed
Context 1 should be closed

**Screenshots**
2022-06-02_14h37_04
2022-06-06_15h00_46

 - Browser Chromium
 - Version 102.0.5005.40

aaltat:
Fixes #2089

### DIFF
--- a/node/playwright-wrapper/interaction.ts
+++ b/node/playwright-wrapper/interaction.ts
@@ -101,8 +101,14 @@ export async function click(
     const selector = request.getSelector();
     const options = request.getOptions();
     const strictMode = request.getStrict();
-    await internalClick(selector, strictMode, options, state);
-    return emptyWithLog(`Clicked element: '${selector}' with options: '${options}'`);
+    const result = await internalClick(selector, strictMode, options, state);
+    if (result) {
+        return emptyWithLog(`Clicked element: '${selector}' with options: '${options}' successfully.`);
+    } else {
+        return emptyWithLog(
+            `Clicked element: '${selector}' with options: '${options}' but silenced Target closed error.`,
+        );
+    }
 }
 
 export async function tab(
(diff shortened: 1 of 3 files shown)


node/playwright-wrapper/interaction.ts
```diff
@@ -117,9 +123,28 @@ export async function tab(
...
+    state: PlaywrightState,
+): Promise<boolean> {
     const locator = await findLocator(state, selector, strictMode, undefined, true);
-    await locator.click(JSON.parse(options));
+    try {
+        await locator.click(JSON.parse(options));
+        return true;
+    } catch (error: unknown) {
+        if (error instanceof Error) {
+            if (error.message.startsWith('locator.click: Target closed')) {
```
Snooz82, line 138: i would think that the two if conditions can be combined by an `and` because in both else branches you throw.

And isn't the `else` unnecessary? Because you can either return if both conditions are met and otherwise just re-raise the error outside the if?!

Snooz82: And i would change the order:

```javascript
 try {
      await locator.click(JSON.parse(options));
      return true;
  } catch (error: unknown) {
      if (!(error instanceof Error && error.message.startsWith('locator.click: Target closed'))) {
           throw error;
      }
      logger.warn('Supress locator.click: Target closed error');
      return false;  
  }

**Review by Snooz82 — COMMENTED** [review]
Some comments

File: node/playwright-wrapper/interaction.ts [hunk 2]
```diff
@@ -101,8 +101,14 @@ export async function click(
...
     const options = request.getOptions();
     const strictMode = request.getStrict();
-    await internalClick(selector, strictMode, options, state);
-    return emptyWithLog(`Clicked element: '${selector}' with options: '${options}'`);
+    const result = await internalClick(selector, strictMode, options, state);
+    if (result) {
+        return emptyWithLog(`Clicked element: '${selector}' with options: '${options}' successfully.`);
+    } else {
+        return emptyWithLog(
+            `Clicked element: '${selector}' with options: '${options}' but silenced Target closed error.`,
```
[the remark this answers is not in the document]
Snooz82: also here the else is irrelevant.
This can be outside the else.
And could we just "conditionally" add the appendix `"Target was closed by click."`

Issue #1098, opened 2021-06-23
[Feature] Promise To Wait For Download with custom timeout argument


adrian-evo:
Currently when I would like to have a custom timeout for this keyword, I would need to use the following workaround:

Suppose the default browser timeout is set globally on suite setup:

```
Suite Setup
    New Context          acceptDownloads=True
    Set Browser Timeout    ${DEFAULT_WAIT}
    ...
```

Then on specific test case, I would need a custom timeout, expecting a longer than normal download timeout:

```
Test Case
    New Page             ${LOGIN_URL}
    Set Browser Timeout    ${LONG_WAIT}
    ${dl_promise}        Promise To Wait For Download    /path/to/download/folder
    Click                \#file_download

    # Restore default timeout if download fails
    ${status}    ${file_obj}    Run Keyword And Ignore Error    Wait For    ${dl_promise}
    Set Browser Timeout    ${DEFAULT_WAIT}
    IF    '${status}' == 'FAIL'
        Take Screenshot
        Fail    ${file_obj}
    END

    File Should Exist    ${file_obj}[saveAs]
    Should Be True       ${file_obj.suggestedFilename}
```

Considering this, it would be useful if _Promise To Wait For Download_ keyword could have a default timeout argument, like e.g. _Wait For Request_ keyword, if possible.

aaltat:
@all-contributors please add @adrian-evo for ideas

PR #3228, opened 2023-11-09
added Timeout to Download and possibility to have multiple downloads …

Snooz82:
…at the same time


Snooz82:
Implements #1098 and more other features

PR #3253, opened 2023-11-17
AutoClosingLevel KEEP

### LINKED ISSUE #1263: Automatic Closing Level MANUAL: keep open after terminating the test
Currently auto_closing_level=MANUAL keeps the browser open during the test execution.
It might be useful however to keep the window even after finishing the test. 

``` diff 
- If automatic closing level is MANUAL, nothing is closed automatically during the test execution is ongoing.
+ If automatic closing level is MANUAL, nothing is closed automatically and the browser is kept open even after the test execution is finished.
```

Using seleniumLibrary

**Additional context**
This might come in handy in RPA settings where RPA tasks no only require human intervention but where RPA prepares a session for the user.

RPA use case:
I would like to open a browser which opens a website and logs me in so that I can use it for what ever my job is.

aaltat:
Will keep browser open also after test execution

Fixes #1263

### DIFF
--- a/Browser/base/librarycomponent.py
+++ b/Browser/base/librarycomponent.py
@@ -26,7 +26,7 @@ from robot.utils import timestr_to_secs
 
 from ..generated.playwright_pb2 import Response
 from ..utils import SettingsStack, get_variable_value, logger
-from ..utils.data_types import DelayedKeyword, HighLightElement
+from ..utils.data_types import AutoClosingLevel, DelayedKeyword, HighLightElement
 
 if TYPE_CHECKING:
     from ..browser import Browser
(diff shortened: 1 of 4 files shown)


Browser/base/librarycomponent.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, library: "Browser") -> None:
     def playwright(self):
         return self.library.playwright
 
+    @property
+    def auto_closing_level(self) -> AutoClosingLevel:
+        return self.library._auto_closing_level
+
```
Snooz82, line 54: Why is in playwright.py this is used and at the other situations the _ one?

aaltat: Need more words?

aaltat: Now I have understood, changed

PR #3419, opened 2024-02-07
Group esbuild esbuild-node-externals for depandabot

aaltat:


aaltat:
Fix based on https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

PR #3509, opened 2024-03-29
Click and Select Options By logging

### LINKED ISSUE #3510: Improve logging done by Click and Select Options By keyword
Click and Select Options By keywords should log the selector and options they are using.

lennartq:
Added info level logging for the Click and Select Options By keywords. In cases where the arguments came from variables, the default log didn't show what was clicked / selected.

Disclaimer: I'm a n00b. Make sure I haven't done something that will crash when something is null or whatever.

Fixes #3510

### DIFF
--- a/Browser/keywords/interaction.py
+++ b/Browser/keywords/interaction.py
@@ -312,7 +312,6 @@ class Interaction(LibraryComponent):
 
         [https://forum.robotframework.org/t//4238|Comment >>]
         """
-        logger.info(f"Clicks the element '{selector}'.")
         self.click_with_options(selector, button)
 
     @keyword(tags=("Setter", "PageContent"))
@@ -364,6 +363,7 @@ class Interaction(LibraryComponent):
 
         [https://forum.robotframework.org/t//5936|Comment >>]
         """
+        logger.info(f"Clicks the element '{selector}'.")
         selector = self.presenter_mode(selector, self.strict_mode)
         with self.playwright.grpc_channel() as stub:
             options = {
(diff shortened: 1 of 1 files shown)


Browser/keywords/interaction.py
```diff
@@ -735,6 +736,7 @@ def select_options_by(
             self.deselect_options(selector)
             return []
 
+        logger.info(f"Selects the option(s) {', '.join(values)} by attribute {attribute} from element {selector}.")
```
aaltat, line 739: I think this breaks the keyword, because values can contains something else than strings.

Browser/keywords/interaction.py
```diff
@@ -312,6 +312,7 @@ def click(self, selector: str, button: MouseButton = MouseButton.left):
 
         [https://forum.robotframework.org/t//4238|Comment >>]
         """
+        logger.info(f"Clicks the element '{selector}'.")
```
aaltat, line 315: I think this logging should go inside of `click_with_options` method. Then it would be same for both keywords.

aaltat (changes requested):

Browser/keywords/interaction.py
lennartq: Done. It could of course log more if it's there but...

Browser/keywords/interaction.py
lennartq: Better?

aaltat: Let’s see what test do say

aaltat (approved):

aaltat:
@allcontributors please add @lennartq for code

PR #3534, opened 2024-04-08
Fix Cookie documentation "expires" instead of "expiry".

### LINKED ISSUE #3532: Documentation: Get Cookies "expiry" value should be "expires" instead
https://marketsquare.github.io/robotframework-browser/Browser.html#Get%20Cookie

Possible typo in the value "expiry" where it should be "expires" instead. See https://playwright.dev/docs/api/class-browsercontext#browsercontextaddcookiescookies

> expiry	Lifetime of a cookie. Returned as datatime object or None if not valid time received.

The typo also appears in the example code

> `Should Be Equal   ${cookie.expiry.year}     ${2020}`

anacomparada:
Fixes #3532


aaltat:
And thank you for your contribution.

PR #3595, opened 2024-05-14
Shrink docker image a little

chrisns:

PR #3604, opened 2024-05-16
Shrink docker image a little

chrisns:
improved (fixed) version of #3595
builds but doesn't test because since its a separate step it'd need to push somewhere.

you'd have to either export the docker image to an artifact and pull it in to the second step, or just merge the two things into a single flow

either-way ci it probably needs refactoring to:
build > test > publish

for ci (non release events) it could also probably be optimized for speed to not bother building non-amd64 images since qemu is slow and it needs to compile things

### DIFF
--- a/.github/workflows/dockers.yml
+++ b/.github/workflows/dockers.yml
@@ -1,6 +1,7 @@
 name: Docker image and tests
 
 on:
+  push:
   release:
     types: [ published ]
   workflow_dispatch:
@@ -17,6 +18,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Login to Docker Hub
+        if: github.event_name != 'push'
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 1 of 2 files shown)


aaltat:
Most of the time in CI, just build only with one time would be enough or great improvement for the current situation. Because functionality is tested in other CI jobs and all docker is just container (not totally true, because there is some custom code that is different/extra from the library) for the library. 

This is the part which is unfamiliar for me, because I am not familiar with Docker and I can’t see which way would be best for the project.

chrisns:
its probably worth checking you can build arm images still when you have dependency updates
theres probably some improvements that could come from caching

aaltat:
Checking arm would be beneficial because in the history it has had it own share of problems. I am more thinking that we don’t have to cover all the possible combinations, just doing one or two is enough.

chrisns:
did you see issues with the execution or just build on ARM?

aaltat:
If my memory serves me well always in build time.  Although I have some nagging feeling that there was some issue with browser binaries between arm and x64 platform, during run time. But I am not sure was it with Docker or with something else.

chrisns:
in which case this PR will mean you'll always have an arm build tested :)
there's further improvements around caching to improve build time and not pulling the latest pypi version and instead building from source, happy to pick up as another PR if we can get this merged now and bedded in?
~c

aaltat:
I am happy with that way.

PR #3628, opened 2024-05-29
Try to fix docker build

aaltat:


joergplewe:
Is it known when it will be pushed to dockerhub? As of today, `latest` is from may16.

aaltat:
There was more fixing needed: https://github.com/MarketSquare/robotframework-browser/pull/3665 Should be there now

PR #3654, opened 2024-06-16
Implementation of Playwright pdf()

### LINKED ISSUE #3481: Add playwright's .pdf() conversion function as keyword
See https://playwright.dev/docs/api/class-page#page-pdf

Supporting the `pdf()` function of playwright would make it possible to test the browser's "print to pdf" functionality. 
Even when screenshots in full length are alreay possible, PDFs represent a different use case. 

Perhaps switching the media type to "print" would also make sense to get an even more realistic result, see https://stackoverflow.com/questions/72324669/playwright-download-via-print-to-pdf 

Happy to hear your thoughts, thanks! :-)

aaltat:
Fixes #3481


aaltat:
Closing on favor of #3660

PR #3758, opened 2024-08-22
feat: only check for node when robot is running

Leemur89:
Hello,

Sometimes the code completion of IDE fails to import the library due to some issues with the PATH, while when running the test the lib is perfectly imported. This can happen when the IDE does not properly initiate the PATH with node, without impacting the run.
This is why I am suggesting to check if node is installed only if the test is actually running, which would allow IDE to load the keywords even if node is not in PATH

### DIFF
--- a/Browser/playwright.py
+++ b/Browser/playwright.py
@@ -30,6 +30,8 @@ from Browser.generated.playwright_pb2 import Request
 from .base import LibraryComponent
 from .utils import AutoClosingLevel
 
+from robot.libraries.BuiltIn import BuiltIn
+
 if TYPE_CHECKING:
     from .browser import Browser
 
@@ -50,7 +52,8 @@ class Playwright(LibraryComponent):
     ):
         LibraryComponent.__init__(self, library)
         self.enable_playwright_debug = enable_playwright_debug
-        self.ensure_node_dependencies()
+        if BuiltIn().robot_running:
+            self.ensure_node_dependencies()
         self.port = str(port) if port else None
         self.playwright_log = playwright_log
 


aaltat:
Could you fix the linting error with ruff.

Leemur89:
Hi @aaltat ,
I've fixed the linting error but now I got an error on the MacOS run with python 3.12

```
Screenshotting With Jpeg Extension And Quality Borders                  
  Traceback (most recent call last):
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    File  /Users/runner/work/robotframework-browser/robotframework-browser/atest/test/02_Content_Keywords/screenshot.robot:49
    T:  Screenshotting With Jpeg Extension And Quality Borders    
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    File  /Users/runner/work/robotframework-browser/robotframework-browser/atest/test/02_Content_Keywords/screenshot.robot:59
      Numbers Are Close    ${size_100}    ${size_101}    10
      |  ${size_100} = 89564 (int)
      |  ${size_101} = 89602 (int)
______________________________________________________________________________
| FAIL |
```

I don't really know what this is about sorry :/

aaltat:
That screenshot test is quite flaky, but seems that reruns did help

aaltat:
This is tricky thing. I understand your need and the change in the code. But we support running library also without Robot Framework and we have unit tests in place for that. Also there is documentation that such feature is supported. 

Now if we make the change, check is not done in when running library without Robot Framework. The library will fail, but in different way. The question is: will that change user experience in significant way and is the easier option to configure IDE in such way that the error doesn’t appear. Example of library does not work in your IDE, does debugging test case, to the library level then work?

aaltat:
@Snooz82 what you think?

Snooz82:
Hi,

No. not that way!
i drafted the topic already and i would just have made a property.
the problem with „robot running“ is, that this would made the usage without robot impossible.

regards

Leemur89:
Hi @aaltat ,

The problem I have is when I launch VSCode from the application list on Mac, it does not properly use my environment variables, and `node` is not in my PATH. Running a test is working fine, but the linter that I'm using fails to find it, hence fails to start the lib and discover the keywords

@Snooz82 would that be better if I do this skip only if robot can be imported?

aaltat:
Configuring Mac with VSCode is possible, I have Mac + VSCode and it works. But that only proves that it works on my machine and I don’t have solution for your problem.

In any case, I don’t see how skipping when Robots Framework can not be imported would solve this problem? Would you care to explain why that would work?

Leemur89:
@Snooz82 what would be the use case of a usage without robot? this is robotframework-browser, I'm actually surprised that this could be used without robot

aaltat:
We discussed in the team and consensus was that we would skip starting the node process if there is not plugins initialized. Because one should  get keywords if plugins are not present, it is just Python  code, but if there are plugins, specially the node extensions, we need the node process to do the discovery.

Snooz82:
This PR should be obsolete due to a different fix in #3822

Issue #3286, opened 2023-11-28
Issues with downloadsPath variable


alpip1997:
**Describe the bug**
It seems that when I try to solve a download promise (set with promise_to_wait_for_download, and resolve with wait_for(promise)) I get a TypeError  because of the variable downloadsPath. This error doesn't happen if I use the version of Robotframework_broser 17.5.2 just with the new version 18.0.0. And also the system must be Linux, in Windows I didn't have the issue. 

**Expected behavior**
The expected behaviu¡our is to solve the promise without any error

aaltat:
I don’t quite see what causes the problem. Would it be possible for you to create an example which also we can run?

alpip1997:
Hello, I cannot give you a version that runs as our code because it is a company website. 
I can give you the lines of code that we use to download. Basically we select the filename, create the promise of downloading and click de button for download it. I dont know if this can be useful.
`filename = name
        if name == "":
            test_case_name = self.get_test_case_name()
            self.stdoutDownloaded += 1
            filename = test_case_name.lower()
            filename= filename.replace(' ', '_')
            filename = f"{filename}_{self.stdoutDownloaded}.stdout"
        
        
        folderPath=self.getEvidenceFolder(self.cmpMode)
        time.sleep(2)
        download_promise = self.playwright.promise_to_wait_for_download(f"{folderPath}/{filename}")
        time.sleep(2)
        self.do_click(xp.download_button)
        self.playwright.wait_for(download_promise)`

 I also want to say that the installation was perform in a computer without internet connection (the website is accesible because the server is set inside a private network). So for the installation I have to donwload de whl files from pypi, manually install the node packages (for the steps rfbrowser init) and give the correct permissions. But this method was perform in the same way for installing the version 17.5 and it works fine, it is just the version 18 and its dependencies which give me this error.

Thanks you for your help.

aaltat:
Can you also send the full stack trace from the error?

aaltat:
@allcontributors please add @alpip1997 for bugs

alpip1997:
Hello, sorry for the delay. We have been having issues with our machines. Recently I reinstaled the version 18 and I got the same error. For the error, the trace that I got in the debug file is:

`================= Original suppressed error =================
TypeError: Cannot read properties of undefined (reading 'downloadsPath')
    at _waitForDownload (/usr/local/lib/python3.10/site-packages/Browser/wrapper/index.js:8835:63)
[...]

PR #3819, opened 2024-10-04
Issue #3286: Handle missing browser options when downloading files

lasselindqvist:
I did not check if this actually solves #3286 but at least it should not make things worse.


aaltat:
Please write a test which prevents us breaking it again.

lasselindqvist:
I'm sorry but I don't have enough expertise with the codebase to do that. If it is a hard requirement, feel free to reject the PR.

aaltat:
Writing an acceptance test is most likely better way to go

aaltat:
Setting up a dev environment is explained in here 
https://github.com/MarketSquare/robotframework-browser/blob/main/CONTRIBUTING.md

PR #3997, opened 2025-01-15
Corrected example comment for Set Geolocation

falcon030:


aaltat:
Fixes #4015

@allcontributors please add @falcon030 for code.

PR #4089, opened 2025-02-28
Add Support for Semantic Selectors

shuhaodo:
# Add Support for Semantic Selectors

## Overview
Implements natural language selectors that remain resilient across DOM changes. This feature allows replacing traditional CSS or XPath selectors with semantic descriptions.

## Examples
```diff
- Click  //span/button[@class="some_class"]
+ Click $"Login Button"
```

## Implementation
This functionality is implemented through AgentQL's Playwright wrapper, which enables prompt-based locators to identify elements based on their purpose rather than their structure.

## Benefits
- **Resilience**: Tests are stable through DOM structure changes
- **Readability**: Test scripts become more intuitive and self-documenting
- **Maintainability**: Reduces the need to update selectors when UI is refactored

## Testing
The example script below uses semantic locators

```robot
*** Settings ***
Library    Browser

*** Test Cases ***
Has Title
    New Page    https://playwright.dev/
    Get Title    contains    Playwright

Get Started Link
    New Page    https://playwright.dev/
    ${element}=    Get Element    $"Get started link"
    Click    ${element}
    ${heading}=    Get Element    $"Installation heading" 
    Get Element States    ${heading}    contains    visible

Get Started Link [Alternative]
    New Page    https://playwright.dev/
    Click    $"Get started link"
    Get Element States    $"Installation header"    contains    visible

Playwright Test
    New Page    https://playwright.dev/
    Get Title    matches   Playwright
    Get Attribute    $"Get started link"    href    ==    /docs/intro
    Click    $"Get started link"
    Get Url    matches    .*intro
```

## Requirements:

- Set the AGENTQL_API_KEY environment variable
- API keys can be created on dev.agentql.com

### DIFF
--- a/node/playwright-wrapper/playwright-invoke.ts
+++ b/node/playwright-wrapper/playwright-invoke.ts
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { configure, wrap } from 'agentql';
 import { Locator, Page } from 'playwright';
 
 import { PlaywrightState } from './playwright-state';
@@ -39,6 +40,11 @@ export async function findLocator(
 ): Promise<Locator> {
     const activePage = state.getActivePage();
     exists(activePage, 'Could not find active page');
+
+    if (selector.startsWith('$"') && selector.endsWith('"')) {
+        return findSmartLocator(activePage, selector.slice(2, -1));
+    }
+
     if (strictMode) {
         selector = selector.replaceAll(' >>> ', ' >> internal:control=enter-frame >> ');
     } else {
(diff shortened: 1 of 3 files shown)


aaltat:
I appreciate your effort, but this type of functionality should be build as an extension for the library and it shouldn’t be part of the library itself. Your code is neatly written and from the changes I can see that writing this type of extension might be difficult in the current architecture we have. Please raise an issue where we can decide what kind of support this type of need would require from the library. Solution should be generic enough that it’s suitable for other types tooling than this one.

As a side note, have you seen https://github.com/robotframework/robotframework/issues/4909

shuhaodo:
It makes sense to create an extension. Noted #4909, we will iterate on the syntax as well.

aaltat:
I feel that making this as an extension is a must. Browser library main focus is to support features of Playwright and therefore we are careful which features and dependencies actually are supported by us.

But your idea is good, basically you want to alter how library searches elements. In practice similar functionality exists in SeleniumLibrary side and I have been wondering would there be similar needs in Browser library side. You are the first one to ask it, so we should think about it.

PR #4114, opened 2025-03-15
Fix ruff checks

rasjani:
* calling os.listdir() to check if subdirectory exists could be slow
 * round() already returns int if ndigits arg is omitted.


aaltat:
Thank you for fixing this one

PR #4150, opened 2025-04-03
implemented highlight_on_failure

Snooz82:
With this it is possible to automatically highlight the selector that caused a failure.
Of course only if the element is there. 

This has to be enabled by the init argument `highlight_on_failure` or by `Set Highlight On Failure` keyword.

If active, during an error, so when the `run_on_failure` keyword is active, the robot variable `${ROBOT_FRAMEWORK_BROWSER_FAILING_SELECTOR}` will contain the failed selector.

So it could also be used for other keywords, like selector healing or so.
@manykarim


aaltat:
Closed on favor of #4197

Issue #4257, opened 2025-06-16
Support `datetime.datetime` in `Add Cookie` argument `expires`


Brownies:
`Get Cookie` returns a dict with a `datetime` expiry, it would be nice to be able to use the returned value directly instead of converting to a string or timestamp in between.

aaltat:
So you want to define possible format of the `expiry` instead of retuning `datatime` object, do I understand you correctly? Do you have some specific formats in your mind, your should the format be user defined?

Brownies:
No, I would like to pass a `datetime` object to `Add Cookie`

aaltat:
Because type hint is a string, then `datatime` object does not work properly? Or is there some sort of error?

Brownies:
Running
```robotframework
*** Settings ***
Library    Browser


*** Test Cases ***
Sample
    New Context
    Add Cookie    cookiename    cookievalue    expires=${{ datetime.datetime.now(datetime.UTC) }}
```
results in a `ValueError: Invalid timestamp '2025-06-16 13:51:07.409012+00:00'.` I'm running Browser 19.5.1, Playwright 10.2.3, RF 7.1.1

The stack trace I get is
```
 	Traceback (most recent call last):
  File "/home/lassi/.cache/pypoetry/virtualenvs/robot-tests-Fa1v0Xsl-py3.12/lib/python3.12/site-packages/Browser/keywords/cookie.py", line 145, in _expiry
    return int(expiry_cleaned)
           ^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '2025-06-1613:51:07'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/lassi/.cache/pypoetry/virtualenvs/robot-tests-Fa1v0Xsl-py3.12/lib/python3.12/site-packages/Browser/browser.py", line 1268, in run_keyword
    return DynamicCore.run_keyword(self, name, args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lassi/.cache/pypoetry/virtualenvs/robot-tests-Fa1v0Xsl-py3.12/lib/python3.12/site-packages/robotlibcore.py", line 173, in run_keyword
    return self.keywords[name](*args, **(kwargs or {}))
[...]

PR #4311, opened 2025-07-27
Support datetime for Add Cookie expiry

Brownies:
For #4257 

Should be fairly simple but do you think more tests are needed?

### DIFF
--- a/Browser/keywords/cookie.py
+++ b/Browser/keywords/cookie.py
@@ -95,7 +95,7 @@ class Cookie(LibraryComponent):
         url: Optional[str] = None,
         domain: Optional[str] = None,
         path: Optional[str] = None,
-        expires: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


aaltat:
Well, what type of test you had in your mind?

aaltat:
Also some unit tests seems fail, could you take a look?

Brownies:
Sure thing, it seems some platforms handle timezones a bit differently.

aaltat:
Yes, timezone is problematic.

Brownies:
Hopefully ok now, I got the workflows enabled on my fork and at least the tests passed there. I also added a test for using the expiry of a cookie as the expiry of another cookie.

aaltat:
CI passed and by a quick glance all looks good from mobile screen. I will take a look later on my laptop screen and see how it looks like.

@allcontributors please add @Brownies for code.

PR #4422, opened 2025-09-25
Interrupt playwright process before killing it

### LINKED ISSUE #4421: Playwright process is not killed after tests are done
**Describe the bug**

Any time I run a test, Playwright leaves behind a running process. This is the same process that logs to `playwight-log.txt`. Because of this, `playwight-log.txt` remains locked (can't be renamed, deleted, written to).

This may be a Windows-only issue. I have no opportunity for testing Linux at this time.

**To Reproduce**

Note: Reproduction may only work on my machine due to the nature of these kinds of things.

1. Run a test
2. Look in task manager
3. During the test, there are 2 node processes
    
    Image

    Note: I have no idea why it creates 2 processes. My best guess is that the first process (the one without icon) is a bootstrapping process. The second process (the one with icon) runs the actual Playwright server code.

4. Let the test finish
5. One of the processes remains and will never exit by itself

    Image

7. `playwright-log.txt` can't be deleted (the remaining process still claims ownership)

**Expected behavior**

All processes exit and the log file is released.

 - OS: Windows 11
 - Any browser
 - `robotframework-browser` 19.8.0
 - `playwright` 1.55.0

**Smartphone (please complete the following information):**
Unclear if affected

**Additional context**

This is very setup-dependent and the information above is pretty sparse. Because of this, I dove into the issue myself. The following change worked for me:

Replace [main/Browser/playwright.py#L221-L224](https://github.com/MarketSquare/robotframework-browser/blob/main/Browser/playwright.py#L221-L224) with the following:

```python
        if playwright_process:
            logger.trace("Closing Playwright process")
            playwright_process.send_signal(signal.CTRL_C_EVENT)
            try:
                playwright_process.wait(1)
            except KeyboardInterrupt:
                logger.trace("Playwright process interrupted")
            except subprocess.TimeoutExpired:
                playwright_process.kill()
                logger.trace("Playwright process killed")
```

Instead of immediately calling `playwright_process.kill()`, try to interrupt the process gracefully first.

This works because on `SIGTERM` (i.e. `.kill()`) Node has no time to close the secondary process. By sending it an interrupt instead, Node has some time to send an interrupt to the secondary process. Thankfully, it needs very little time to do so. In this code, we wait for up to 1 second for the Node processes to close. If it takes them longer, fall back to the current method of calling `.kill()`.

Lakitna:
Closes #4421


aaltat:
@allcontributors please add @Lakitna for code

aaltat:
Looks like there is problem with Mac, but it also looks that it should be easy to overcome.

Lakitna:
Same with Linux.. The signal I had to use only exist in Windows.

It looks like simply replacing `signal.CTRL_C_EVENT` with `signal.SIGINT`, but that does not work on Windows... I'll have to spin up a Linux VM to see if `SIGINT` works there.

aaltat:
We do provide GitPod https://github.com/MarketSquare/robotframework-browser/blob/main/.gitpod.yml and last time I looked, it’s free for non commercial use

Lakitna:
I had a bunch of issues, but got a basic Linux setup working. That should've be simple 🤷‍♂️ VirtualBox was being weird. Gitpod wanted a credit card (maybe due to linked orgs in my Github account?) Stuff like that.

Anyway. Long story short: Sending `SIGINT` works great on Linux!

I may have an option tomorrow to borrow someones Mac to double check MacOs too.

aaltat:
You could just push it and let the CI do the check for you. Most of the time, I will use one OS (Mac in my case) for development and I trust CI to run rest of the OS/Python/NodeJS combinations.

Lakitna:
For any other change I would do the same. But testing this particular thing also requires me to check if 

- any processes where left behind after the run
- it can successfully close the process with only the first signal

These things are much easier to check in this, admittedly convoluted, way.

Lakitna:
Confirmed that MacOS works with `SIGINT`.

aaltat:
There seems to error when running unit test in Windows 
```
- generated xml file: D:\a\robotframework-browser\robotframework-browser\utest\output\pytest_xunit.xml -
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! KeyboardInterrupt !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\grpc\_channel.py:2089: KeyboardInterrupt
```

Perhaps mock will solve this problem.

aaltat:
I looked at the problem and I do not see good way to solve the problem with signal. I also looked how pabot is doing it: https://github.com/mkorpela/pabot/blob/main/src/pabot/pabot.py#L616-L617

```python
process.terminate()
process.wait()
```

Have you tried would it solve also your problem?

Lakitna:
[According to the docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill), `.kill()` is the same as `.terminate()` on Windows. Because... Windows I guess.

I've never checked if Pabots process stops properly. I never noticed it doing so. But it's less noticeable since it doesn't lock any files. It may also not have the same thing that Node does with the secondary process since it's a Python process.

aaltat:
Because sending the signal.CTRL_C_EVENT is problematic, should we try to catch the exception in Windows and then kill it?

Lakitna:
I had a similar thought. I tried to catch it every way in the test code, but could not catch it in any way. 

Turns out that the issue is not directly with the new way of closing the process. It indirectly causes `atexit` registration of `cleanup_numbered_dir`. This one appears to be registered by Pytest to clean up a temp dir. Because of this, I can't find a way to wrap it in try-except.

Working with the utests also made me realize that the node processes created by the utests are never terminated with the old code. I see the same behavior that made me open the PR in the first place. With the new code (this branch) they are terminated. 

So, I'm stuck in a situation where the new code is an improvement, but it will break utests in a way that I can't figure out how to fix. The only "solution" I've come up with is to mess with the atexit registered functions. But that really is not great.

aaltat:
Well, I am happy to skip the utest on windows that causes problem with pytest. If it works leave comment why it’s done.

Lakitna:
I've traced the issue to 2 utests. Both of them manually call `Playwright.close()`. That seems to cause the issue somehow. 

- `test_playwright_exit_handler`
- `test_playwright_double_close`

Without these 2 tests, it passes on my machine. This commit skips those 2 tests on Windows.

aaltat:
Ah atexit shoots now us in the foot. Perhaps call https://docs.python.org/3/library/atexit.html#atexit.unregister in windows helps in Pytest side.

aaltat:
Another solution came up in my mind. We know the process ID of the node process. So if `process.kill()` doesn’t work, then we could kill with `psutil`. Would that be possible solution for you? I am happy to add psutil for a dependency

Lakitna:
Oh that may work. psutil has `children(recursive=False)` which should allow us to kill all children of the Playwright process (bootstrap process and runtime process). Then I can remove the whole `CTRL_C_EVENT` and `SIGINT` thing!

Lakitna:
I have high hopes for this solution!

Lakitna:
Working with the list of child processes, I also found out why node spins up 2 processes. It's caused by Volta. I use Volta to manage the NodeJS versions on my machines. The first process was `Volta/node.exe` which in turn opens `NodeJS_{version}/node.exe`.

This explains a lot. It explains why not everyone has this issue. It explains why there are 2 processes. It even explains why one of the processes has an icon in task manager while the other does not.

Lakitna:
I have fixed the type lint issue. 
But I have no idea why some pipeline agents can't find `psutils` despite installing it successfully.

Somehow there is also an agent that needs `rfbrowser init`. No idea how that happened either

Lakitna:
I did not expect a secondary list of dependencies at `./Browser/dev-requirements.txt`. I think adding `pyutils` there solves the issues

Lakitna:
I stupidly added `psutils` to `pyproject.toml` instead of `psutil` (without `s`). I have done smarter things in my life.

~Please let it pass checks now~

aaltat:
No worries, I didn’t see it either and I also do same kind of mistakes.

Lakitna:
All checks have passed! 🎉🎉

With that I'd say it's ready to merge unless you have any review comments

aaltat:
Great work and thank you for seeing it trough all hoops.

aaltat:
> Working with the list of child processes, I also found out why node spins up 2 processes. It's caused by Volta. I use Volta to manage the NodeJS versions on my machines. The first process was `Volta/node.exe` which in turn opens `NodeJS_{version}/node.exe`.

Would you willing to give [BrowserBatteries](https://pypi.org/project/robotframework-browser-batteries/) a try and see do you encounter similar problems. I am trying to understand do I need to implement similar thing for BrowserBatteries.

Lakitna:
I tried to do that, but could not find a great way of installing the playwright binaries. What command do you expect people to run when they get this message?

```
╔═════════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright Test or Playwright was just installed or updated. ║
║ Please run the following command to download new browsers:              ║
║                                                                         ║
║     npx playwright install                                              ║
║                                                                         ║
║ <3 Playwright Team                                                      ║
╚═════════════════════════════════════════════════════════════════════════╝
```

aaltat:
My bad, I should have pointed you to correct release https://pypi.org/project/robotframework-browser-batteries/19.9.0b3/ You also need Beta3 from Browser library https://pypi.org/project/robotframework-browser/19.9.0b3/ to be installed.

Lakitna:
I do have both of those installed. I installed `robotframework-browser-batteries=19.9.0b3` with Poetry which also pulled in `robotframework-browser=19.9.0b3`. So that part works nice and easy :)

I made a test that opens Chrome to about:blank and ran it.

This is where I got the error I shared before. Here is some extra error log:

```
{"level":30,"time":"2025-09-30T18:42:25.586Z","pid":7444,"hostname":"DESKTOP-S3HAJ4Q","msg":"Listening on 127.0.0.1:54117"}
{"level":30,"time":"2025-09-30T18:42:25.606Z","pid":7444,"hostname":"DESKTOP-S3HAJ4Q","msg":"Start of node method newBrowser"}
{"level":30,"time":"2025-09-30T18:42:25.612Z","pid":7444,"hostname":"DESKTOP-S3HAJ4Q","msg":"Error of node method  newBrowser"}
================= Original suppressed error =================
browserType.launch: Executable doesn't exist at C:\code\scratch\robotframework-tmp\.venv\lib\site-packages\Browser\wrapper\node_modules\playwright-core\.local-browsers\chromium_headless_shell-1193\chrome-win\headless_shell.exe
╔═════════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright Test or Playwright was just installed or updated. ║
║ Please run the following command to download new browsers:              ║
║                                                                         ║
║     npx playwright install                                              ║
║                                                                         ║
║ <3 Playwright Team                                                      ║
╚═════════════════════════════════════════════════════════════════════════╝
    at _newBrowser (C:\snapshot\robotframework-browser\Browser\wrapper\index.js:12365:54)
    at newBrowser (C:\snapshot\robotframework-browser\Browser\wrapper\index.js:12887:33)
    at PlaywrightServer.newBrowser (C:\snapshot\robotframework-browser\Browser\wrapper\index.js:13456:34)
    at Object.onReceiveHalfClose (C:\snapshot\robotframework-browser\node_modules\@grpc\grpc-js\build\src\server.js:1464:25)
    at BaseServerInterceptingCall.maybePushNextMessage (C:\snapshot\robotframework-browser\node_modules\@grpc\grpc-js\build\src\server-interceptors.js:595:31)
    at BaseServerInterceptingCall.handleEndEvent (C:\snapshot\robotframework-browser\node_modules\@grpc\grpc-js\build\src\server-interceptors.js:635:14)
    at ServerHttp2Stream.<anonymous> (C:\snapshot\robotframework-browser\node_modules\@grpc\grpc-js\build\src\server-interceptors.js:394:18) {
  name: 'Error'
}
```

When I look at `.venv\lib\site-packages\Browser\wrapper`, there is no `node_modules`. 

image

Because of this I assumed that I'm missing a command.

aaltat:
Ah, then you need to give also ‘rfbrowser install` command to install Playwright browser binaries. It should populate the Browser\wrapper folder.

Lakitna:
With batteries, the issue does not happen. I'm not sure why, but the second node process does not appear:

image

It looks like you don't need to worry (at least right now) :)

aaltat:
That confirms my assumption. With the BrowserBatteries we don’t use NodeJS, instead there is a self contained binary. Thank you for trying it out and providing feedback.

PR #4493, opened 2025-10-16
Lazy loading of logger tuning

### LINKED ISSUE #4491: Make logger initialization lazy to avoid permission errors
My production server's file system is locked down. In the upgrade from 19.8.0 to 19.10.0, I now get:

[ ERROR ] Error in file '/srv/.../common.resource' on line 6: Importing library 'Browser' failed: PermissionError: [Errno 13] Permission denied: '/srv/app/rfbrowser.log'

This PR fixes the PermissionError by implementing lazy logger initialization

In v19.10.0, the logger in Browser/entry/constant.py is initialized at module import time (lines 33-53), which causes PermissionError in production environments where (introduced in https://github.com/MarketSquare/robotframework-browser/pull/4329). The logger attempts to create a log file in:

    The package directory which is read-only (common in Docker containers)
    The current working directory which is read-only
    And the application runs as a non-root user which limits writes overall.

The logger initialization was triggered by a new import added in v19.10.0 (commit https://github.com/MarketSquare/robotframework-browser/commit/4088d384013856833afaffc4ec8af3d3c8bcea1f): Browser/utils/misc.py now imports PLAYWRIGHT_BROWSERS_PATH from Browser/entry/constant, which is part of the standard library import chain. This means the logger setup code runs whenever the Browser library is imported, even if CLI logging is never needed.

This commit implements lazy initialization by:

    Moving logger setup into an _init_logger() function
    Only calling _init_logger() when log() is actually used
    Updating coverage_combine.py to use the lazy logger

The logger is now only created when CLI commands need it, not during normal library imports.

Goal: Fixes permission errors in read-only filesystem environments while maintaining full compatibility with existing CLI functionality.

aaltat:
Fixes #4491

### DIFF
--- a/Browser/entry/constant.py
+++ b/Browser/entry/constant.py
@@ -16,6 +16,7 @@ import os
 import platform
 import re
 import sys
+from functools import cached_property
 from logging.handlers import RotatingFileHandler
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


aaltat:
I made some refactoring @databyte would you able to give it a try so that we are sure that it still fixes your problem?

PR #4545, opened 2025-11-18
Fix promise_to_upload_file hanging issue (#4546)

### LINKED ISSUE #4546: Mirod adjustements to promise_to_upload_file keyword
**Describe the bug**
Add missing return type annotation (-> Future) to promise_to_upload_file
Add return statement to _upload_file method to return response.log

luchoagomezt:
- Add missing return type annotation (-> Future) to promise_to_upload_file
- Add return statement to _upload_file method to return response.log
- Fixes issue where robot.exe would not finish execution after using promise_to_upload_file
- Makes _upload_file consistent with other promise methods like _wait_for_download

Fixes #4546


aaltat:
It looks that there is hickup with GitHub and it can not connect to npm registry. 

@allcontributors please add @luchoagomezt fir code.

PR #4640, opened 2026-01-22
feat: add Secret support (Closes #4269)

### LINKED ISSUE #4269: Add support for Robot Framework 7.4 Secret type
Add support for Robot Framework 7.4 Secret type. For Type and Fill Secret keyword.

Noordsestern:
Closes #4269 

The more I think about it, the more corners I imagine where this won't work (do we have dictionaries of secrets, too??).

Anyways, if we have only str values: As far as I see there is 1 place where to choose between Secret or "default" behaviour. 

Making a difference between Robot versions in the code looks ugly. I saw it back in the days when Robot supported Python 2. Probably won't work after Robot 10, but until then 7.3 is probably out of support anyway :D 

I tested utest with:
- Robot Framework 7.3.2 (expectation: works although Secret type does not exist) 🍏 
- Robot Framework 7.4.1 (expectation: work when Secret type exist) 🍏 

However, I was not sure how to execute pytest cases based on the Robot version. So, I did not add new pytests for Secret type, yet.

EDIT: I have seen that the pipelines never test with different Robot Framework versions. Maybe that was just never necessary, yet. How should the pytests be set up then? Should there be version specific test cases or only tests compatible with latest Robot Framework version?

Thought I first get your feedback.


aaltat:
Hi, I will try to look this soon, but some robocon deadlines are lurking and I need to do those things first. Sorry for the delay

aaltat:
> The more I think about it, the more corners I imagine where this won't work (do we have dictionaries of secrets, too??).

Yes there are dictionaries and dictionaries can have secret in it and it should just work. 

> Anyways, if we have only str values: As far as I see there is 1 place where to choose between Secret or "default" behaviour.

This PR alone is not enough,  need to define `Secret` also in the keywords type annotations (because without the annotation, RF will convert the Secret to str and stuff does not work as expected.) Because we support older versions of RF (but we do not test it in the CI), we need to make our own implementation of `Secret` until 7.4 is the minimum version.

Because now we have need to test with different RF version, we should do it also in CI. 

> However, I was not sure how to execute pytest cases based on the Robot version. So, I did not add new pytests for Secret type, yet. 

Pytest has skipif decorator https://docs.pytest.org/en/stable/how-to/skipping.html#id1 which is handy for this type of things.

Noordsestern:
> need to define `Secret` also in the keywords type annotations (because without the annotation, RF will convert the Secret to str and stuff does not work as expected.) 

😱 I had done this in my local fix, but somehow not in the PR. I wonder why it still worked then... i think, i must have forgotten that i made fixes and robot test cases in my venv, but then forgot to port the tests and annotations to the PR 😞 

> Because we support older versions of RF (but we do not test it in the CI), we need to make our own implementation of `Secret` until 7.4 is the minimum version.

But how would users of RF <7.4 define a Secret?

> Because now we have need to test with different RF version, we should do it also in CI.

👍🏻  doubles the amount of test shards, though 😬 but there is no other way.

> Pytest has skipif decorator https://docs.pytest.org/en/stable/how-to/skipping.html#id1 which is handy for this type of things.

✅

aaltat:
This is much bigger than seen by first glimpse, because this is not only about users defining secret. It is **mostly** about the keywords argument types. Browser library relies heavily on Robot Framework [automatic argument conversion](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#argument-conversion). Now example [Fill Secret](https://marketsquare.github.io/robotframework-browser/Browser.html#Fill%20Secret) keyword `secret` argument type is now `str`. In Python code it looks like:
```python
def fill_secret(self, selector: str, secret: str, force: bool = False):
...
```

 When Support for Robot Framework [Secret](https://robot-framework.readthedocs.io/en/master/autodoc/robot.utils.html#robot.utils.secret.Secret) is made, keywords `secret` argument will be union of `str` and `Secret`. In Python code it looks like:
```python
def fill_secret(self, selector: str, secret: str | Secret, force: bool = False):
...
```
And the above must work regardless what Robot Framework version they have installed. Because we can not ship different Browser library versions, based on what Robot Framework version they have installed. 

Most likely we need
1. User RF `Secret if RF version is 7.4 or greater in type hints.
1. Use Browser library Secret if if RF version is less than 7.4 in type hints. Most likely this is copy paste code from RF 7.4 but we need to do try/catch or if/else when importing it.
1. Write test for every keyword using secret and pass in real secret object. There should be test for str type already in place. 
1. Change every keyword type hint, which uses secret, from `secret: str` to `secret: str | Secret`. Also we need to take in secret in dictionaries. 
1. Enhance CI to run test with RF 7.4 and 6.1. We might need to some skipping if test is targeted for different RF version less then 7.4.

aaltat:
Closing this and making secret support in #4790 4790

Issue #4644, opened 2026-01-26
Get Element Count shows incorrect error when count is not expected


johndoorzon:
**Describe the bug**
When I want to check if an element is _not_ on the page, I use:
Get Element Count    >locator<    ==    0
This now gives me a wrong error when the locator can be found:

_Element count for selector <locator> (int) should be '0.0' (float)_

I.m.o. a count should always be an int.
When the count is as expected, no error is returned, so happy flow is correct.

**To Reproduce**
Run the test in attached robot file
[test.robot.txt](https://github.com/user-attachments/files/24856386/test.robot.txt)

**Expected behavior**
A correct error that does not expect a float, but an int

**Desktop (please complete the following information):**
 - OS: Any (tested on Windows and Linux)
 - Browser chromium
 - robotframework: 7.4.1
 - robotframework-browser: 19.12.3

aaltat:
This is true and should be easy to fix. Want to make PR to fix it?

johndoorzon:
Yes, I will make a pull request

PR #4645, opened 2026-01-26
Fix for issue #4644

johndoorzon:
Get Element Count returns an int, so int_dict_verify_assertion should be used.


aaltat:
There seems to be some errors, can you take a look?

aaltat:
This does not work, because the problem is actually in assertion engine and it does not have proper support for int. 

Therefore closing this PR

aaltat:
I fixed assertion engine. See #4652 for updated in Browser.

PR #4693, opened 2026-02-27
Add Electron application support

shenthils-ui:


shenthils-ui:
Three new Browser library keywords for testing Electron desktop apps:

- `New Electron Application` — launches an Electron `.exe` and attaches
  Playwright to its main window. Automatically skips splash screens
  (windows with no title) and waits for the real main window.
- `Wait For Electron App Ready` — waits for `domcontentloaded` and
  returns the window title, confirming the Browser library gRPC channel
  is live and all page-level keywords are ready.
- `Close Electron Application` — cleanly closes the Electron process
  and removes it from the Browser library state stack.

atest/test/01_Browser_Management/electron.robot
aaltat: Test must be written in a way that they are always run in CI, electron support will not be an exception. This means, 

1. No skipping. We can skip some OS if building electron app is hard in some OS. But generally skipping all the time is not accepted.
2. Source of code for  the electron app which is used to test the library must exist in this project.
3. Building of the electron app must be possible in CI and locally 
4. Build test app must be used in CI dynamically. Commit build electron app in this repo is not accepted. 
5. Documentation must be added for how electron app works, build and so on.

Browser/keywords/electron.py
```diff
@@ -0,0 +1,154 @@
...
+            response = stub.LaunchElectron(req)
+            logger.info(response.log)
+            return (
+                response.browserId,
+                response.id,
+                NewPageDetails(page_id=response.pageId, video_path=None),
+            )
+
+    @keyword(tags=("Wait", "BrowserControl"))
+    def wait_for_electron_app_ready(
```
aaltat: Do we need this a separate keyword, should launch do this by default?

Browser/keywords/electron.py
```diff
@@ -0,0 +1,154 @@
...
+        self,
+        state: str = "domcontentloaded",
```
aaltat: We use Enums to define constants. I think one already exist for this usage, see https://github.com/MarketSquare/robotframework-browser/blob/main/Browser/utils/data_types.py

Browser/wrapper/index.js
aaltat: This should be part of the PR. This always generated on the fly.

node/playwright-wrapper/generated/playwright_grpc_pb.js
aaltat: The files in `generated` folder must not be part of the PR. They are generated on the fly, remove them.

package.json
```diff
@@ -42,6 +42,7 @@
     "monocart-coverage-reports": "^2.12.9",
     "pino": "^10.3.1",
     "playwright": "^1.58.2",
+    "protoc-gen-js": "^3.21.4-4",
```
aaltat: Why this is neede? I did not see imports using it?

shenthils-ui:
Closing this PR and reopening a cleaner version with only the required source changes, tests, and docs based on the review comments.

Issue #4391, opened 2025-09-10
grpc message size limits causes keyword failures with large DOMs


frste:
**Describe the bug**
In case of a website dom which contains a lot of content, keywords like "Get Page Source" may fail with message: "CLIENT: Received message larger than max (4200443 vs. 4194304)"

**To Reproduce**
Unfortunately I can't provide a public website for testing this behavior

**Expected behavior**
The called keywords should not fail because of reduced message length limits

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: chromium
 - Version 140.0.3485.54

**Additional context**
- The root cause for this problem seems to be the limits of the gprc channel. When setting appropriate max size for the gprc_channel solves the problem for this specific case
- `grpc.insecure_channel(f"{self.host}:{self.port}", options=[
        ('grpc.max_message_length', 4300000),
        ('grpc.max_send_message_length', 4300000),
        ('grpc.max_receive_message_length', 4300000),
    ],)`

Could you please check this case and either adjust the grpc_channel settings to appropriate values for handling large websites (we also have seen cases of even bigger dom content) or make these values configurable.

Kind regards, Frank

aaltat:
No need to create example, I know what causes that error. GRPC has max limit which it can transfer with one message. The fix is to move to use stream with GRPC messaging model

@allcontributors please add @frste for bugs

aaltat:
Can you provide a PR to fix the problem?

frste:
@aaltat I’m not yet familiar with the concept, but I’ll give it a try

aaltat:
Wait For Response has a stream https://github.com/MarketSquare/robotframework-browser/blob/main/protobuf/playwright.proto#L542 it can serve as an example how protobuf and Python + NodeJS code should be changed.

PR #4729, opened 2026-03-13
Handle large chunked JSON responses across the Browser gRPC boundary

tonyr-qf:
Previously, large responses could exceed the effective gRPC payload boundary. This was especially relevant for:
- large page source payloads via `Get Page Source` (#4391)
- extension keyword results returned via `CallExtensionKeyword`

This PR makes those response-side cases deterministic and adds regression coverage.

Note that `CallExtensionKeyword` request arguments can still suffer from the same limitation in the request direction.
I have left it unchanged intentionally, as large argument payloads in extension keywords are generally avoidable through design choices.


aaltat:
There is a lint error in atest. But I do not this time agree with robocop and your version of the code is more readable. Could add comment to line, to suppress the error for robocop?

Browser/browser.py
```diff
@@ -1103,11 +1111,19 @@ def call_js_keyword(self, keyword_name: str, **args) -> Any:
...
             for response in responses:
                 logger.info(response.log)
-            if response.json == "":
+                if response.bodyPart:
+                    body = body + response.bodyPart
+                if response.json:
+                    last_json = response.json
+            if body != "":
+                return json.loads(body)
+            if last_json == "":
```
aaltat: Use: 
```python
if not last_json:
```

Fix also 1078

aaltat:
I looked why this fails, we missed one logging which was used in run_on_failure suite to check that correct keyword is called. While fixing that, I remembered that similar chunking logic was also done somewhere in the code. I found it and refactored it to use new chunking logic. Therefore I created https://github.com/MarketSquare/robotframework-browser/pull/4743 and I am going to close this PR. Do not worry, I will fully credit you in the release notes and thank you for the effort in fixing the issue (I just fixed few minor things).

PR #4737, opened 2026-03-16
fix #4735 and avoid github references in package.json

### LINKED ISSUE #4735: avoid github references in package.json
We run into build-problems due to github references in package.json. Please avoid such references:

https://github.com/MarketSquare/robotframework-browser/blob/main/package.json#L15

`"@yao-pkg/pkg": "github:yao-pkg/pkg"`

should be like

` "@yao-pkg/pkg": "^6.14.1"`

DirkRichter:
avoid github references in package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "@types/react-dom": "^19.2.3",
     "@types/react-router-dom": "^5.3.3",
     "@types/strip-comments": "^2.0.4",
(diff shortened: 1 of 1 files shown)


aaltat:
@allcontributors please add @DirkRichter for code

aaltat:
Thanks for fixing this problem

PR #4991, opened 2026-06-23
fix: Get Text returns option text instead of value for select elements

### LINKED ISSUE #4953: Get Text Keyword returns value attribute instead of the text of an option element
**Describe the bug**
I have noticed a behavior change in "Get Text" Keyword from version 19.12.0 to 20.0.0

**To Reproduce**
sample_adress_form.html

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="UTF-8" />
        <title>City Selection</title>
    </head>
    <body>
        <div class="city-field">
            <label for="city">City</label>
            <select id="city" name="city">
                <option value="value_1">Berlin</option>
            </select>
        </div>
    </body>
</html>
``` 

potetntial_bug.robot
```robotframework
*** Settings ***
Library     Browser  


*** Test Cases ***
Select Berlin From Dropdown
    [Documentation]    Opens the sample page and selects 'Berlin' from the city dropdown.
    New Browser  chromium  headless=False
    New Page  ${CURDIR}${/}sample_adress_form.html
    ${city_text}  Get Text    select#city  
    Should Be Equal   ${city_text}  Berlin    
    Close Browser`
``` 

**Expected behavior**
With Version 19.12.0 this test passes 
With Version 20.0.0  we get the error message "value_1 != Berlin"

**Screenshots**

Image

**Desktop
 - OS: Windows Windows 11 Pro
 - Browser Chrome for Testing
 - Version 148.0.7778.96 (Official Build) (64-bit)

**Additional context**
Slack Channel message https://robotframework.slack.com/archives/C015KB1QSDN/p1781012172157479?thread_ts=1780896883.081299&cid=C015KB1QSDN

mvanhorn:
## What

Restores the pre-v20 behavior of `Get Text` on a `<select>` element so it returns the visible option text instead of the selected option's `value` attribute.

## Why

A regression landed between Browser library v19.12.0 and v20.0.0: `Get Text` on a `<select>` started returning the selected option's `value` attribute (e.g. `value_1`) instead of its visible text (e.g. `Berlin`). The cause is in `node/playwright-wrapper/getters.ts`: commit 6b5fe6b expanded the tag check in `_getTextContentNoTextType` from `INPUT`/`TEXTAREA` to also include `SELECT`, routing selects through `locator.inputValue()` (the value attribute) instead of `innerText()`.

In v19.12.0 a `<select>` fell through to `innerText()`. This PR removes `SELECT` from the `inputValue` branch so selects fall through to `innerText()` again, matching the original behavior. The explicit `text_type=inputValue` path is unchanged for callers who still want the value attribute.

## Changes

- `node/playwright-wrapper/getters.ts`: narrow the `inputValue` branch in `_getTextContentNoTextType` back to `TEXTAREA`/`INPUT` only; update the fallback log message; add a comment documenting the intentional `<select>` fall-through.
- `node/playwright-wrapper/__tests__/getters.test.ts`: update the `SELECT` no-text-type unit test to expect `innerText` instead of `inputValue`.
- `utest/test_get_text.py`: add a unit test documenting that select option text is returned unchanged through the keyword wrapper.
- `atest/test/02_Content_Keywords/basic_getters_get_text.robot`: update the select acceptance assertions to check the visible option text rather than the value attribute.

Fixes #4953

### DIFF
--- a/node/playwright-wrapper/getters.ts
+++ b/node/playwright-wrapper/getters.ts
@@ -118,12 +118,13 @@ enum TextType {
 async function _getTextContentNoTextType(locator: Locator): Promise<string[]> {
     logger.info(`Getting text content without text type`);
     const tag = await locator.evaluate((e) => e.tagName);
-    if (tag === 'SELECT' || tag === 'TEXTAREA' || tag === 'INPUT') {
+    // <select> intentionally falls through to innerText() to restore pre-v20 (v19.12.0) behavior per issue #4953.
+    if (tag === 'TEXTAREA' || tag === 'INPUT') {
         logger.info(`Element is ${tag}, get inputValue`);
         const inputValue = await locator.inputValue();
         return [inputValue];
     } else {
-        logger.info(`Locator was not an <input>, <textarea>, or <select> element, falling back to innerText.`);
+        logger.info(`Locator was not an <input> or <textarea> element, falling back to innerText.`);
         const innerText = await locator.innerText();
         return [innerText];
     }
--- a/utest/test_get_text.py
+++ b/utest/test_get_text.py
@@ -8,8 +8,20 @@ class Response:
     items = ["element text"]
 
 
+class SelectResponse:
+    log = "Log text"
+    items = ["Dog"]
+
+
 def test_get_text(ctx: MagicMock):
     getter = Getters(ctx)
     getter._get_text = MagicMock(return_value=Response())  # type: ignore[assignment]
     text = getter.get_text("//div")
     assert text == "element text"
+
+
+def test_get_text_select_returns_option_text(ctx: MagicMock):
+    getter = Getters(ctx)
+    getter._get_text = MagicMock(return_value=SelectResponse())  # type: ignore[assignment]
+    text = getter.get_text("id=pet-select")
+    assert text == "Dog"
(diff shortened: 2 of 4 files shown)


utest/test_get_text.py
```diff
@@ -8,8 +8,20 @@ class Response:
...
+
+
 def test_get_text(ctx: MagicMock):
     getter = Getters(ctx)
     getter._get_text = MagicMock(return_value=Response())  # type: ignore[assignment]
     text = getter.get_text("//div")
     assert text == "element text"
+
+
+def test_get_text_select_returns_option_text(ctx: MagicMock):
```
aaltat, line 23: Remove this test, it is not needed

node/playwright-wrapper/getters.ts
```diff
@@ -118,12 +118,13 @@ enum TextType {
 async function _getTextContentNoTextType(locator: Locator): Promise<string[]> {
     logger.info(`Getting text content without text type`);
     const tag = await locator.evaluate((e) => e.tagName);
-    if (tag === 'SELECT' || tag === 'TEXTAREA' || tag === 'INPUT') {
+    // <select> intentionally falls through to innerText() to restore pre-v20 (v19.12.0) behavior per issue #4953.
```
aaltat, line 121: Remove comment, this is documented in tests

aaltat (changes requested):

aaltat:
@allcontributors please add @mvanhorn for code

aaltat:
Cleanup done in https://github.com/MarketSquare/robotframework-browser/pulls

mvanhorn:
Thanks for the merge and the cleanup pass, @aaltat.

mvanhorn:
Thanks @aaltat! Get Text returning the option text instead of the value should match what users expect.

PR #5155, opened 2026-08-15
Recognise the library's own keywords by identity, not by the name “Browser”

### LINKED ISSUE #5154: Keyword call banner is silently disabled when Browser is imported with AS alias or subclassed
Three listener hooks compare the library name against the literal string `"Browser"`:

* `Browser/browser.py:935` — `or attrs["libname"] != "Browser"` in `_start_keyword`, gating `self._show_keyword_call(attrs)`
* `Browser/browser.py:939` — `if "secret" in attrs["kwname"].lower() and attrs["libname"] == "Browser":` → `self._set_logging(False)`
* `Browser/browser.py:1031` — the same condition in `_end_keyword`, restoring the log level

The listener itself is registered as `self` (`Browser/browser.py:509`), so the hooks do fire in these cases — only the name comparison fails.

That comparison is false in two ordinary situations:

1. `Library    Browser    AS    PW`
2. a user library built by subclassing `Browser`, imported under its own name

### 1. Keyword-call banner is silently a no-op — confirmed

The banner is a `<style id="kwCallBanner">` element injected into `<head>` (`KW_CALL_BANNER_FUNCTION`, `Browser/browser.py:109`). Same suite, same resource file, `show_keyword_call_banner=True`, real `robot` runs, webkit headless, probed with `Get Element Count    css=style#kwCallBanner`:

| Import | Banner element count |
| --- | --- |
| `Library    Browser` | **1** — `textContent` starts `body::before {\n    content: 'Get Property    css=style#kwCallBanner    textContent';` |
| `Library    Browser    AS    PW` | **0** |
| `Library    MyLib` where `class MyLib(Browser)` | **0** |

The element is never created. Nothing appears in the page, in videos, in traces, or in the log, and no warning is issued — `show_keyword_call_banner=True` and `Enable Presenter Mode` simply do nothing for anyone using an alias or a subclass.

### 2. Secret log suppression is skipped — but no secret was disclosed

The `_set_logging` suppression at `:939` / `:1031` genuinely does not run for those imports. Observed in `output.xml` for the same three keywords and arguments:

```
Library Browser:   -- Fill Secret owner=Browser  args=['id=pw', '$PWD']    (no messages logged)
Library MyLib:     -- Fill Secret owner=MyLib    args=['id=pw', '$PWD']
                        [ TRACE ] Arguments: [ 'id=pw' | '$PWD' ]
                        [ TRACE ] Presenter mode: False, selector: id=pw, strict: True
                        [ TRACE ] Return: None
```

**The secret value itself never appeared in clear text in either run.** Grepping `output.xml` and `log.html` for both secret values, at `TRACE:INFO`, gave 0 occurrences in every configuration, including a forced failure path (`Fill Secret` against a missing selector, so the timeout error and the on-failure screenshot were exercised).

Three independent defences already keep the value out: the Robot Framework argument is the unresolved literal `$PWD` / `%ENV` (resolution happens inside `resolve_secret`, `Browser/base/librarycomponent.py:243`), an RF 7.4 `Secret` renders as `Secret(value=<secret>)`, and `fill_secret` / `type_secret` pass `log_response=False` (`Browser/keywords/interaction.py:846`). `_set_logging` is belt-and-braces on top of those.

So this half is a consistency and robustness problem rather than a disclosure: any future code path that does log a resolved secret would be unprotected for aliased and subclassed imports. Worth noting that losing the suppression also changes the failure output in the other direction — plain `Browser` swallows the timeout traceback entirely, while the aliased/subclassed import shows it.

### Suggested fix

Compare identity rather than the name — resolve the library instance for `attrs["libname"]` and check `isinstance(instance, Browser)`, or compare against the instance's own resolved name — so aliases and subclasses behave like `Browser`. `Browser/keywords/waiter.py:299-304` already does the `isinstance` form for `Wait For Condition` and can serve as the pattern.

### Environment

Browser 20.3.0 (`3b3ec7a9`), Robot Framework 7.4.1, Python 3.14.7, macOS, webkit.

sujeito-operator:
Fixes #5154.

## What changes

The three listener hooks stop asking *"is this keyword called `Browser`?"* and start
asking *"is this keyword mine?"* — `attrs["libname"]` is resolved to a library instance
and compared with `self`:

* `_start_keyword` — the keyword-call banner gate (`browser.py:935`)
* `_start_keyword` — the secret log suppression (`browser.py:939`)
* `_end_keyword` — its restore (`browser.py:1031`)

The lookup goes through `BuiltIn().get_library_instance(all=True)` and is cached per
library name, because these hooks fire on **every** keyword in the suite, not only on
ours. The cache is cleared in `_start_suite`, since a name only resolves to an instance
within one namespace. The condition order in `_start_keyword` is changed so the two
cheap boolean tests short-circuit before the resolution — the operands are pure, so the
meaning is unchanged.

## Reproduced first, on the unpatched tree

Four real `robot` runs. No browser is needed for this: the listener hooks fire *before*
the keyword body, so a keyword that then fails for want of a browser still records the
decision. `_show_keyword_call` and `_set_logging` were patched to record their caller
and arguments; everything else is the library as shipped.

| Suite | Import | before | after |
| --- | --- | --- | --- |
| `plain` | `Library Browser` | banner + suppress/restore | unchanged |
| `alias` | `Library Browser AS PW` | **nothing at all** | banner + suppress/restore |
| `subclass` | `Library mylib.py` where `class MyLib(Browser)` | **nothing at all** | banner + suppress/restore |
| `both` | `Browser` **and** `MyLib` together | see below | one instance per keyword |

That last row is worth a second look, because it is a defect in the *current* behaviour
rather than a missing one. With both imported, `Browser.Click` produced **two**
`_show_keyword_call` calls — one of them from the `MyLib` instance, which does not own
that keyword and may not even have a page open — while `MyLib.Click` produced none.

## The one place this departs from the suggested fix, and it was measured

The issue suggests `isinstance(instance, Browser)`. I ran both, changing only that line.

Both instances are registered as listeners (`ROBOT_LIBRARY_LISTENER = self`) and both are
notified of every keyword, so under `isinstance` each one claims the other's keywords.
The banner half of that is cosmetic. `_set_logging` is not: it stashes the previous level
on `self` but sets a **global** one, so around a single `Fill Secret` the two instances
interleave —

```
isinstance   Browser  suppress -> global NONE, stashed INFO
             MyLib    suppress -> global NONE, stashed NONE   <-- stashes the clamp
             Browser  restore  -> global INFO
             MyLib    restore  -> global NONE                 <-- and puts it back
```

— and the run is left at `NONE`. In the probe the `Log    after the secret keyword`
that follows is present in `output.xml` as a `<msg>` element under identity and absent
under `isinstance`. Identity gives one listener per keyword and no interleaving.

If you would rather have `isinstance` anyway, it is one line in `_is_own_keyword` and I
am happy to change it — but the `_set_logging` interleaving would want handling
separately.

## Tests

`utest/test_listener_own_keyword.py`, 14 tests. Ten cover `_is_own_keyword` directly
(own name, alias, a name that is not ours, another library, subclass, two instances not
claiming each other, caching in both directions, the cache clearing, and
`RobotNotRunningError` — which is unreachable from a listener hook and defensive only).
Four drive `_start_keyword`/`_end_keyword` and assert what they decided; **two of those
four fail on `main`** and are the actual regression guard.

## What I did not run, and what I checked instead

* **`inv atest` was not run.** It needs a node build and Playwright browsers; this
  machine has neither. **Please treat CI as the authority**, and note that I have
  therefore *not* seen a `<style id="kwCallBanner">` element appear in a real page under
  an aliased import — I have only seen `_show_keyword_call` reached with the right
  instance. The issue's own `Get Element Count` measurement is the check I could not
  repeat.
* `inv utest` on this tree: **13 failed, 298 passed, 2 skipped, 23 errors** — and
  **identical** before and after the patch apart from my 14 new passes. Every failure is
  `Couldn't find node` or a missing `node/dynamic-test-app` build.
* `ruff format --check`, `ruff check`: clean on both changed files; the repo-wide counts
  (40 findings in other `utest` files under ruff 0.16.3) are unchanged by this patch.
* `mypy --config-file Browser/mypy.ini Browser/`: **22 errors before, 22 after**. They
  are all `playwright_pb2` stub and `BrowserBatteries` artefacts of a hand-generated
  protobuf on this box. The `# type: ignore[attr-defined]` on the
  `get_library_instance` result mirrors the one already on `waiter.py:299`.

Built and verified against `cd1f8544a2`.

### DIFF
--- a/Browser/browser.py
+++ b/Browser/browser.py
@@ -30,7 +30,7 @@ from assertionengine import AssertionOperator
 from overrides import overrides
 from robot.api.deco import library
 from robot.errors import DataError
-from robot.libraries.BuiltIn import EXECUTION_CONTEXTS, BuiltIn
+from robot.libraries.BuiltIn import EXECUTION_CONTEXTS, BuiltIn, RobotNotRunningError
 from robot.running.arguments import PythonArgumentParser
 from robot.utils import secs_to_timestr, timestr_to_secs
 from robot.utils.robottypes import is_falsy
@@ -508,6 +508,7 @@ class Browser(DynamicCore):
         self.ROBOT_LIBRARY_LISTENER = self
         self.scope_stack: dict = {}
         self.suite_ids: dict[str, None] = {}
+        self._own_libname_cache: dict[str, bool] = {}
         self.current_test_id: str | None = None
         self._rf_context = _RFContextTracker()
         self._playwright_state: PlaywrightState = PlaywrightState(self)
@@ -860,7 +861,32 @@ def {name}(self, {", ".join(argument_names_and_default_values_texts)}):
     def coverage_output(self) -> Path:
         return self.browser_output / "coverage"
 
+    def _is_own_keyword(self, libname: str) -> bool:
+        """Is ``libname`` the name *this* instance is imported under?
+
+        The listener is registered as ``self`` (``ROBOT_LIBRARY_LISTENER = self``), so the
+        listener hooks see every keyword in the suite and have to recognise their own.
+        Comparing ``libname`` against the literal ``"Browser"`` gets that wrong whenever
+        the library is imported ``AS`` another name or subclassed into a user library.
+
+        Identity is used rather than ``isinstance``: when a suite imports both ``Browser``
+        and a subclass of it, *both* instances are registered as listeners and both are
+        notified of every keyword, so an ``isinstance`` test would make each of them react
+        to the other's keywords.
+        """
+        cached = self._own_libname_cache.get(libname)
+        if cached is None:
+            try:
+                instances = BuiltIn().get_library_instance(all=True)
+            except RobotNotRunningError:
+                return False
+            cached = instances.get(libname) is self  # type: ignore[attr-defined]
+            self._own_libname_cache[libname] = cached
+        return cached
+
     def _start_suite(self, name, attrs):
+        # A name resolves to an instance within one namespace; a new suite is a new one.
+        self._own_libname_cache.clear()
         self.suite_ids[attrs["id"]] = None
         self._add_to_scope_stack(attrs["id"], Scope.Suite)
         self._rf_context.start_suite(attrs["id"], attrs.get("longname", name))
@@ -930,11 +956,13 @@ def {name}(self, {", ".join(argument_names_and_default_values_texts)}):
         if not (
             self.show_keyword_call_banner is False
             or (self.show_keyword_call_banner is None and not self.presenter_mode)
-            or attrs["libname"] != "Browser"
             or attrs["status"] == "NOT RUN"
+            or not self._is_own_keyword(attrs["libname"])
         ):
             self._show_keyword_call(attrs)
-        if "secret" in attrs["kwname"].lower() and attrs["libname"] == "Browser":
+        if "secret" in attrs["kwname"].lower() and self._is_own_keyword(
+            attrs["libname"]
+        ):
             self._set_logging(False)
 
         if attrs["type"] == "Teardown":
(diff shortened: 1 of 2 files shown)


aaltat:
I am not sure about the issue and issue needs more thinking before it can proceed to implementation. Closing the PR.

PR #5182, opened 2026-08-24
docs: fix typo commiting -> committing

vaibhav8a:
One-word typo fix in `CORE_TEAM_AGREEMENT.md:39`: "We agree on all of this by commiting..." -> "committing".


aaltat:
Yes, this has been there for there 6 years and it is not user facing documentation. It’s agreement with contributors, well to behave and act nicely. Although your fix is technically correct, I am going not to allow it, because it will mess the history of the signed agreement.

PR #5238, opened 2026-09-11
fix: Retry a screenshot Chromium refuses outright

### LINKED ISSUE #5237: Take Screenshot fails on Chromium's transient 'Unable to capture screenshot' instead of retrying the capture
`Take Screenshot` fails the keyword when Chromium answers the capture with:

```
Error: page.screenshot: Protocol error (Page.captureScreenshot): Unable to capture screenshot
Call log:
  - taking page screenshot
  - waiting for fonts to load...
  - fonts loaded
```

Chromium refuses that one capture and takes the very same picture when asked again a moment later, so the keyword is reporting a transient browser hiccup as a test failure.

### Evidence

Found in this project's own CI, all history (301 runs, 838k results, 2026-06-07 → 2026-09-11):

* **21 occurrences** across 20 runs and 18 distinct commits, spread evenly from 2026-06-10 onwards. Not a regression — it is as old as the data.
* **One leg only**: `macos-latest`, Python 3.11, node 24.x, shard 1/4 — 21 failures in 287 runs (~7%). The same suite on `windows-latest` (291 + 192 runs) and `ubuntu-latest` (180 + 69 runs) never failed, so it is not node 24 or a Playwright version on its own.
* **Always the first capture after a cold browser start.** `atest/test/02_Content_Keywords/screenshot.robot` takes eight `fullPage=True` screenshots and ran 278 times on another macOS leg with zero failures of this kind.
* **Not a timeout** — the failing calls returned in 363–1412 ms, mostly faster than the passing median of 782 ms.
* **In 21 of 21 cases a capture on the same page succeeded a few hundred milliseconds later**: `keyword_error` runs the `run_on_failure` screenshot before re-raising, and every one of those logged `Screenshot successfully captured to: .../fail-screenshot-1.png`. None logged `could not be run on failure`.

Playwright (1.63.0) does not special-case this error, so it propagates to the keyword.

### Effect on users

Besides an explicit `Take Screenshot` failing, the same hiccup during the automatic `run_on_failure` screenshot is swallowed by `keyword_error` and logged as `Keyword '...' could not be run on failure:` — an unexplained warning attached to an unrelated real failure.

### Status

**The library is not changed.** A retry inside `takeScreenshot` was written and rejected on review: it makes the library paper over a browser refusal that only one CI leg has ever shown, and no user has reported it.

What was done instead is test side only. `atest/test/00_Clean_Output_Dir/01_initial_import.robot` forces a rendered frame before the capture that was failing — a double `requestAnimationFrame`, which is what Playwright itself does before the screenshots it compares (`rafrafTimeout`, on the `expectScreenshot` path) but not on the `page.screenshot` path this library uses.

So a user meeting the same refusal still sees `Take Screenshot` fail. This issue stays as the record of that, and of the evidence above, in case it turns up outside CI.

aaltat:
Chromium intermittently answers Page.captureScreenshot with `Protocol error (Page.captureScreenshot): Unable to capture screenshot` and then takes the very same picture when asked again, so Take Screenshot was reporting a transient browser hiccup as a keyword failure.

takeScreenshot now retries once, and only on that error. Retrying a timeout here would quietly cost the caller twice the timeout they asked for, so the retry waits 250 ms, then runs inside what is left of the original budget rather than a fresh one, and is skipped when too little is left to be worth spending - with timeout 0 read as Playwright reads it, no timeout at all. If the retry fails too, the caller is told the browser refused the capture rather than that a timeout they never configured expired.

Fixes: #5237

### DIFF
--- a/atest/test/00_Clean_Output_Dir/01_initial_import.robot
+++ b/atest/test/00_Clean_Output_Dir/01_initial_import.robot
@@ -16,6 +16,8 @@ Take Screenshot
     File Should Not Exist    ${OUTPUT DIR}/browser/screenshot/initial_screenshot.png
     File Should Not Exist    ${OUTPUT DIR}/browser/screenshot/second_screenshot.png
     File Should Not Exist    ${OUTPUT DIR}/browser/screenshot/third_screenshot.png
+    # Wait For Elements State done because without it the first capture is fail randomly on mac
+    Wait For Elements State    ${SELECTOR_PREFIX_SPACED}id=table1    stable
     ${initial_screenshot} =    Take Screenshot    initial_screenshot    fullPage=True
     VAR    ${initial_screenshot} =    ${initial_screenshot}    scope=GLOBAL
     File Should Exist    ${initial_screenshot}
--- a/atest/test/00_Clean_Output_Dir/02_second_import.robot
+++ b/atest/test/00_Clean_Output_Dir/02_second_import.robot
@@ -1,7 +1,11 @@
 *** Settings ***
-Library     Browser    retry_assertions_for=4 sec
-Library     OperatingSystem
-Resource    ../variables.resource
+Documentation
+...              If Variable \${initial_screenshot} not found, it means 01 failed before it could
+...              publish the path as global variable. Nothing is wrong in this file; look at 01.
+
+Library          Browser    retry_assertions_for=4 sec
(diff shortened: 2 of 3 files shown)


aaltat:
Didn’t do the retrying, instead made the test to wait to element in the page

